Issue #18, opened 2020-11-04
Add support for parallel batch processing messages


astubbs:
Currently the interface only provides the user a way to process single messages in parallel. There are use cases however where the user may want to processes batches of messages in parallel as well.

For example, say the API they're calling out to actually supports batches of requests, for example Elastic Search batch API that can take several documents at once.

The API might look like: 
```
void pollBatch(int batchLevel, Consumer<List<ConsumerRecord<K, V>>> usersVoidConsumptionFunction);
```
So the user could request batches of 5 records, which could then be transformed into a single HTTP request. These batches of 5 messages would all be run in parallel, up to the usual concurrency limits. The same could be added to the vertx module.

Further from this, options could be added to specify requirements for the batch - for example, give me 5 messages in a batch that all have the same key, are all in the same partition etc.

astubbs:
POC https://github.com/confluentinc/parallel-consumer/pull/19

raphaelauv:
This feature is absolutely necessary for my use case

And since the official Kafka consumer lib is bulk I think this wrapper should support it.

ismarslomic:
I can also confirm that this feature is highly wanted. I´m working as we speak with a use case where we need to move records from Kafka to Elasticsearch and doing this in bulks would reduce high volume of REST API calls

raphaelauv:
this will not be part of release 1.0.0 ?

astubbs:
@raphaelauv yes I think it should be. I'm currently working on another project atm, but hope to get back to PC 1.0 release soon.

astubbs:
FYI ok, there’s a couple test failures still - i think they need updating. But the batching feature seems feature complete now (lol). Need to implement some test for the vertx and reactor versions, but should have this merged early next week.

astubbs:
> this will not be part of release 1.0.0 ?

Merged

PR #19, opened 2020-11-04
feature: #18 Add simple parallel batching support

astubbs:
Changes the engine to always use lists of records, where non batch processing is the case where the batch size is one.


parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumer.java
```diff
@@ -54,6 +55,8 @@
      */
     void poll(Consumer<ConsumerRecord<K, V>> usersVoidConsumptionFunction);
```
mhowlett, line 56: FWIW my opinion of what the API could like is pretty much formulated here (and i like how it turned out):

https://github.com/mhowlett/howlett-kafka-extensions/blob/master/src/Howlett.Kafka.Extensions/Streaming/AsyncProcessor.cs

is it expected that the user will ever change the subscription after the processor is started? if not, should the subscription be specified on construction?

I don't like that the thing passed to the poll method is called a consumer, and the poll method is called poll. "registering" a "handler" would be better terminology.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumer.java
```diff
@@ -54,6 +55,8 @@
...
 
+    void pollBatch(int batchLevel, Consumer<List<ConsumerRecord<K, V>>> usersVoidConsumptionFunction);
```
mhowlett, line 58: for extensibility, this would better be in a config object, not a separate parameter.
the single message handler case may need this as well.

i would probably overload the method name (what you call it poll) for both the single and batch cases. i.e don't have pollBatch as well.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelEoSStreamProcessor.java
```diff
@@ -116,6 +117,11 @@
      */
     private Optional<ConsumerRebalanceListener> usersConsumerRebalanceListener = Optional.empty();
 
+    /**
+     * The number of messages to attempt pass into the {@link #pollBatch} user function
+     */
+    private int batchLevel = 5;
```
mhowlett, line 123: batch size would be a better name imo

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumer.java
astubbs: This is due to the Javiers construct for a consuming function being called a `Consumer` - I agree the naming isn't great. We could wrap the java `Consumer`, in order to rename it, but that would limit it's flexibility as all functions would have to inherit from out renamed Consumer.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumer.java
astubbs: Unfortunately due to type erasure, we can't have unlimited overloading, if the only difference is the generics type (as this erased and the compiler can't tell the difference).

How does a batch level apply to a single record consumer function? Or are you suggesting there is no single record consumer function, only a consumer which could take a list of records of size one?

ismarslomic:
I think documentation should clarify relation between the Consumer property `max.poll.records` and the argument `batchLevel` of `pollbatch()`. Should we set both or only the latter? If both, should they be set to same value?

In addition it would be useful if documentation clarifies how error handling and offset commits are handled in batch processing

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumer.java
JorgenRingen, line 58: I think I would consider moving `specifiedBatchLevel/batchSize` from method signature to `ParallelConsumerOptions`. I think it's more natural there, possible to default, and makes it easier to add more "batch-related" config in the future (min number of records before callback invoked for example).

astubbs: mmh very good point. consider it done! any opinion on defaults?

astubbs:
> I think documentation should clarify relation between the Consumer property `max.poll.records` and the argument `batchLevel` of `pollbatch()`. Should we set both or only the latter? If both, should they be set to same value?

good suggestion! Will do so - hold me to it.

> In addition it would be useful if documentation clarifies how error handling and offset commits are handled in batch processing

what sort of errors are you thinking about?

Off the top of my head, by default, if there is an error processing a batch of messages, the same thing would happen as if for a single message - the messages will all be returned to the internal queues, and will be retried after the default delay expires.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumer.java
JorgenRingen: Hmm, unsure tbh! Maybe something like `KafkaConsumer#max.poll.records / ParallelConsumerOptions#maxConcurrency` would ensure some kind of distribution among the threads.

Not having a default and throwing IllegalStateException if `pollBatch` is called without specifying batchSize is also a viable option imo so that the user is forced to make a choice.

ismarslomic:
> what sort of errors are you thinking about?
> 
> Off the top of my head, by default, if there is an error processing a batch of messages, the same thing would happen as if for a single message - the messages will all be returned to the internal queues, and will be retried after the default delay expires

I was not thinking on specific errors, but in general it would be useful to have a section in docs that states exactly what you are saying above, so users are aware of their responsibilities in this aspect!

New feature request - "**Retry settings**"
It would be nice to have some sort of flexibility to configure the built in retry mechanisms in `pollBatch()`, for example `enableRetry`, `maxRetryAttempts` and `waitDuration`. Then you can either configure it for your needs or choose to solve it yourself.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumer.java
astubbs: > is it expected that the user will ever change the subscription after the processor is started? if not, should the subscription be specified on construction?

Yeah this is a good point, and I've wrestled with it. One advantage of not having it in the constructor, is it leaves it open for `registering` other sorts of handlers after construction. It's certainly fathomable that this could happen. (of course, see your other comment).

astubbs: > FWIW my opinion of what the API could like is pretty much formulated here (and i like how it turned out):

Unfortunately my C-foo isn't up to scratch to really grok what the api is there :/

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumer.java
astubbs: > for extensibility, batchLevel would better be in a config object, not a separate parameter.

yeah I think we have unanimous agreement on that

astubbs:
> New feature request - "**Retry settings**"
> It would be nice to have some sort of flexibility to configure the built in retry mechanisms in `pollBatch()`, for example `enableRetry`, `maxRetryAttempts` and `waitDuration`. Then you can either configure it for your needs or choose to solve it yourself.

Thanks again for your input! We should track this in a different feature, it's something I've been planning. I've created: https://github.com/confluentinc/parallel-consumer/issues/65 Enhanced retry epic. Let's continue the conversation there, and let me know if you have any further thoughts! I'll do something basic for the next version (enable/disable and then skip or die probably)...

astubbs:
@ismarslomic @JorgenRingen do you have any ordering (or relative ordering) expectations or requirements for the messages within the batches?

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumer.java
astubbs: >KafkaConsumer#max.poll.records / ParallelConsumerOptions#maxConcurrency 
hmm, unfortunately it's not as simple as that, as the records are pipelined heavily, with a loading factor proportional to the consumption rate by the system. Given that, I think any number I hard code would be totally arbitrary, and really is heavily if not 100% dependent on the users consuming function use case.. so I think forcing the user to choose with the exception you suggest is the way to go..

astubbs:
ok, implementation is ready for testing. I'm not entirely satisfied with the interface, but let me know what you think. I.e. we could make batching the only interface, and have the default batch size be == 1. But then the user function would always need to be a consumer which takes a list of records, which is kind of odd. Depends if you think batching would be the more common use case, or if single record processing would be..

astubbs:
Does anyone have time to test this branch out? If not that's ok, I'll release it as 3.1.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelEoSStreamProcessor.java
```diff
@@ -835,6 +844,31 @@ private void processWorkCompleteMailBox() {
...
+            usersVoidConsumptionFunction.accept(recordList);
+            return UniLists.of(); // user function returns no produce records, so we satisfy our api
+        };
+        Consumer<Object> voidCallBack = (ignore) -> log.trace("Void callback applied.");
+        supervisorLoop(wrappedUserFunc, voidCallBack);
+    }
+
+    private void validateBatch() {
+        if (!options.getBatchSize().isPresent()) {
+            throw new IllegalArgumentException("Using batching function, but no batch size specified in options");
```
JorgenRingen, line 862: Perhaps this message should be a bit clearer and refer to ParallelConsumerOptions explicitly? For new users coming from kafka-consumer, `pollBatch` would be the default/natural way to process messages as kafkaConsumer.poll is only batch-based. Most users don't read docs very carefully, so this exception will probably be the first encounter with PC for many users, and batch size might be confused with `max.poll.records` etc.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumerOptions.java
```diff
@@ -131,6 +132,27 @@
...
+     * <p>
+     * Note that there is no relationship between the Consumer setting of {@code max.poll.records} and this configured
+     * batch size, as this library introduces a large layer of indirection between the managed consumer, and the managed
+     * queues we use. This indirection effectively disconnects the processing of messages from "polling" them from the
+     * managed client, as we do not wait to process them before calling poll again. We simply call poll as much as we
+     * need to in order to keep our queues full with enough work to satisfy demand - and if we have enough then we
+     * actively mange pausing our subscription so that we can continue calling {@code poll} without pulling in even more
+     * messages.
+     */
+    private final Integer batchSize;
```
JorgenRingen, line 147: Great doc! 👍

FrankR85:
Do you already when this feature will be merged? Thanks!

1123:
Getting the following exception when running tests on this branch:

```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 1.761 s <<< FAILURE! - in io.confluent.parallelconsumer.examples.streams.StreamsAppTest
[ERROR] io.confluent.parallelconsumer.examples.streams.StreamsAppTest  Time elapsed: 1.761 s  <<< ERROR!
java.lang.ExceptionInInitializerError
Caused by: org.testcontainers.containers.ContainerLaunchException: Container startup failed
Caused by: org.testcontainers.containers.ContainerFetchException: Can't get Docker image: RemoteDockerImage(imageName=confluentinc/cp-kafka:latest, imagePullPolicy=DefaultPullPolicy())
Caused by: com.github.dockerjava.api.exception.NotFoundException:
Status 404: {"message":"No such image: testcontainers/ryuk:0.3.0"}
```

Tests are running fine on the master branch.

1123:
I have reapplied the changes of this pull requests to a new branch based on the most recent version of the master branch here: https://github.com/1123/parallel-consumer/tree/batching3 . This new branch also includes a new test for ensuring that the average observed batch size does not deviate too much from the configured batch size, and two new configuration options for the initial dynamic load factor as well as a delay for the control loop. I am sure there are some things that need to be reviewed and reworked. So any feedback is appreciated.

astubbs:
ok @1123 your update to the PR looks good! I've given you access, can you force push to this repo and features/batching (the branch in this PR), and we'll get about getting it merged!

1123:
Hey Antony, thanks for reviewing the changes. I force pushed the changes. I will have to revert two things though (distribution management in the pom.xml, and the change to the artifact version in parallel-consumer-core and parent). Also batching feature in the vertx and the reactor modules is still missing.

build-parallel-consumer-core-without-tests.sh
```diff
@@ -0,0 +1,2 @@
+export JAVA_HOME=$(/usr/libexec/java_home -v13)
```
astubbs, line 1: move these scripts into bin/

parallel-consumer-core/pom.xml
```diff
@@ -8,7 +8,7 @@
     <parent>
         <groupId>io.confluent.parallelconsumer</groupId>
         <artifactId>parallel-consumer-parent</artifactId>
-        <version>0.4.0.2-SNAPSHOT</version>
+        <version>0.4.0.2-batching-2</version>
```
astubbs: reminder: revert

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumerOptions.java
```diff
@@ -109,6 +109,26 @@
...
+    /**
+     * When using batching, the received batch size often is considerably lower than the requested batch size.
+     * This may be a problem when a certain batch size is expected for communication with an external system
+     * (e.g. a Web Service that works best when sending 50 events per request).
+     * This parameter allows slowing down the processor thread, such that enough messages are received to closely
+     * fulfill the requested batch size.
+     */
+
+    @Builder.Default
+    public final Integer processorDelayMs = 0;
```
astubbs, line 130: reminder: investigate

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/DynamicLoadFactor.java
```diff
@@ -69,6 +71,18 @@
     private long lastSteppedFactor = currentFactor;
     private Instant lastStepTime = Instant.MIN;
 
+    /**
+     * When using batching, the default initial dynamic load factor 2 is often too low.
+     * Therefore, this constructor allows to set an initial dynamic load factor via the options.
+     * It may make sense to set the initial dynamic load factor to 2 * batchsize.
+     */
+
```
astubbs, line 79: formatting

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/DynamicLoadFactor.java
```diff
@@ -69,6 +71,18 @@
...
+    public DynamicLoadFactor(ParallelConsumerOptions options) {
+        this.currentFactor = options.getInitialDynamicLoadFactor() == null ?
```
astubbs, line 81: investigate utility

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/WorkManager.java
```diff
@@ -225,6 +224,16 @@ private boolean maybeRegisterNewRecordAsWork(final ConsumerRecord<K, V> rec) {
      */
     // todo refactor
     public List<WorkContainer<K, V>> maybeGetWork(int requestedMaxWorkToRetrieve) {
+        // TODO: this is just a temporary solution to allow for more messages to arrive before batches are created.
+        // This ensures an average batch size that is more close to the requested batch size, since this method will
+        // return more work. Probably this should be refactored to be part of the method "tryToEnsureAvailableCapacity".
+        // TODO: Also we should only wait in case not enough records are present.
+        try {
+            Thread.sleep(getOptions().getProcessorDelayMs());
```
astubbs, line 232: reminder: investigate / remove

parallel-consumer-core/src/test/java/io/confluent/parallelconsumer/BatchTest.java
```diff
@@ -0,0 +1,110 @@
...
+import static org.awaitility.Awaitility.waitAtMost;
+
+/**
+ * Basic tests for batch processing functionality
+ */
+@Slf4j
+public class BatchTest extends ParallelEoSStreamProcessorTestBase {
+
+    @Test
+    void averageBatchSizeTest() {
```
astubbs, line 30: reminder: investigate

astubbs (commented):
step

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/WorkManager.java
```diff
@@ -189,17 +192,26 @@ private void ingestPolledRecordsIntoQueues(final int requestedMaxWorkToRetrieve)
      */
     // todo refactor - move into it's own class perhaps
     public List<WorkContainer<K, V>> maybeGetWorkIfAvailable(int requestedMaxWorkToRetrieve) {
+//        // TODO: this is just a temporary solution to allow for more messages to arrive before batches are created.
```
astubbs, line 195: remove

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/DynamicLoadFactor.java
```diff
@@ -68,6 +69,17 @@
     private int currentFactor = DEFAULT_INITIAL_LOADING_FACTOR;
     private long lastSteppedFactor = currentFactor;
     private Instant lastStepTime = Instant.MIN;
+//
```
astubbs, line 72: remove

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumerOptions.java
```diff
@@ -107,6 +109,24 @@
         PERIODIC_CONSUMER_ASYNCHRONOUS
 
     }
+    // todo remove
```
astubbs, line 112: remove

parallel-consumer-core/src/test/java/io/confluent/parallelconsumer/BatchTest.java
```diff
@@ -0,0 +1,148 @@
...
+        int processorDelayMs = 30;
+        int initialDynamicLoadFactor = targetBatchSize * 100;
+
+        //
+        super.setupParallelConsumerInstance(ParallelConsumerOptions.builder()
+                .batchSize(targetBatchSize)
+                .ordering(ParallelConsumerOptions.ProcessingOrder.UNORDERED)
+                .maxConcurrency(maxConcurrency)
+//                .processorDelayMs(processorDelayMs)
+//                .initialDynamicLoadFactor(initialDynamicLoadFactor)
```
astubbs, line 45: remove

parallel-consumer-core/src/test/java/io/confluent/parallelconsumer/BatchTest.java
```diff
@@ -0,0 +1,148 @@
...
+        var numRecords = new AtomicInteger(0);
+        long start = System.currentTimeMillis();
+        RateLimiter statusLogger = new RateLimiter(1);
+
+        parallelConsumer.pollBatch(recordList -> {
+            int mytarget = targetBatchSize;
+            int size = recordList.size();
+
+            try {
+//                log.info("Batch size {}", size);
```
astubbs, line 65: remove

astubbs (commented):
great! ;)

astubbs:
ok, there’s a couple test failures still - i think they need updating. But the batching feature seems feature complete now. Need to implement some test for the vertx and reactor versions, but should have this merged early next week.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelEoSStreamProcessor.java
```diff
@@ -25,10 +23,11 @@
 public class ParallelEoSStreamProcessor<K, V> extends AbstractParallelEoSStreamProcessor<K, V>
         implements ParallelStreamProcessor<K, V> {
 
-    /**
-     * The number of messages to attempt pass into the {@link #pollBatch} user function
-     */
-    private int batchLevel = 5;
+// todo remove
```
astubbs, line 26: remove

parallel-consumer-core/src/test/java/io/confluent/parallelconsumer/AbstractParallelEoSStreamProcessorTestBase.java
```diff
@@ -205,8 +204,9 @@ protected void setupParallelConsumerInstance(ParallelConsumerOptions parallelCon
 
         subscribeParallelConsumerAndMockConsumerTo(INPUT_TOPIC);
 
-        parentParallelConsumer.setLongPollTimeout(ofMillis(DEFAULT_BROKER_POLL_FREQUENCY_MS));
-        parentParallelConsumer.setTimeBetweenCommits(ofMillis(DEFAULT_COMMIT_INTERVAL_MAX_MS));
+        // todo shouldn't need if test harness is also properly event driven?
```
astubbs, line 207: remove?

parallel-consumer-core/src/test/java/io/confluent/parallelconsumer/BatchTestMethods.java
```diff
@@ -0,0 +1,147 @@
...
+
+import static io.confluent.parallelconsumer.ParallelConsumerOptions.ProcessingOrder.PARTITION;
+import static io.confluent.parallelconsumer.ParallelConsumerOptions.ProcessingOrder.UNORDERED;
+import static java.time.Duration.ofSeconds;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.waitAtMost;
+
+@Slf4j
+@RequiredArgsConstructor
+public abstract class BatchTestMethods<POLL_RETURN> {
```
astubbs, line 24: docs

parallel-consumer-core/src/test/java/io/confluent/parallelconsumer/BatchTestMethods.java
```diff
@@ -0,0 +1,147 @@
...
+
+    //todo rename
+    public void averageBatchSizeTestMethod(int numRecsExpected) {
```
astubbs, line 27: name methods better

parallel-consumer-reactor/src/test/java/io/confluent/parallelconsumer/reactor/CopiedBatchTest.java
```diff
@@ -0,0 +1,154 @@
...
+import static io.confluent.csid.utils.StringUtils.msg;
+import static io.confluent.parallelconsumer.ParallelConsumerOptions.ProcessingOrder.PARTITION;
+import static java.time.Duration.ofSeconds;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.waitAtMost;
+
+/**
+ * Basic tests for batch processing functionality
+ */
+//tood delete
```
astubbs, line 32: delete

parallel-consumer-reactor/src/test/java/io/confluent/parallelconsumer/reactor/ReactorBatchTest.java
```diff
@@ -0,0 +1,88 @@
...
+            protected void averageBatchSizePoll(AtomicInteger numBatches, AtomicInteger numRecords, RateLimiter statusLogger) {
+                reactorPC.reactBatch(recordList -> {
+                    return pollInner(numBatches, numRecords, statusLogger, recordList);
+                });
+            }
+
+            @Override
+            public void setupParallelConsumer(int targetBatchSize, int maxConcurrency, ParallelConsumerOptions.ProcessingOrder ordering) {
+                //
+                ParallelConsumerOptions<Object, Object> options = ParallelConsumerOptions.builder()
```
astubbs, line 53: passthrough

parallel-consumer-vertx/src/test/java/io/confluent/parallelconsumer/vertx/VertxBatchTest.java
```diff
@@ -0,0 +1,154 @@
...
+        batchTestMethods = new BatchTestMethods<>() {
+
+            @Override
+            protected KafkaTestUtils getKtu() {
+                return ktu;
+            }
+
+            @SneakyThrows
+            @Override
+            protected Future<String> pollStep(List<ConsumerRecord<String, String>> recordList) {
```
astubbs, line 48: clean up

pom.xml
```diff
@@ -366,6 +366,34 @@
...
+            <dependency>
+                <groupId>me.tongfei</groupId>
+                <artifactId>progressbar</artifactId>
+                <version>0.9.0</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>com.google.flogger</groupId>
+                <artifactId>flogger</artifactId>
+                <version>0.6</version>
```
astubbs, line 378: extract

astubbs (commented):
minor

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -191,13 +194,15 @@ public AbstractParallelEoSStreamProcessor(ParallelConsumerOptions newOptions) {
         options = newOptions;
         options.validate();
 
+//        this.dynamicExtraLoadFactor = new DynamicLoadFactor(options);
```
astubbs, line 197:
```suggestion
```

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -839,19 +928,46 @@ private void commitOffsetsMaybe() {
...
         }
+//        else {
+//            if (log.isDebugEnabled()) {
+//                if (wm.hasCommittableOffsets()) {
+//                    log.debug("Have offsets to commit, but not enough time elapsed ({}), waiting for at least {}...", getTimeSinceLastCommitAttempt(), timeBetweenCommits);
+//                } else {
+//                    log.trace("Could commit now, but no offsets committable");
+//                }
+//            }
+//        }
```
astubbs, line 940: needs investigation

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/DynamicLoadFactor.java
```diff
@@ -19,6 +19,7 @@
  */
 // todo make so can be fractional like 50% - this is because some systems need a fractional factor, like 1.1 or 1.2 rather than 2
 @Slf4j
+//@NoArgsConstructor
```
astubbs, line 22:
```suggestion
```

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ShardManager.java
```diff
@@ -84,16 +89,22 @@ Object computeShardKey(ConsumerRecord<K, V> rec) {
...
      * @return Work ready in the processing shards, awaiting selection as work to do
      */
-    public int getWorkQueuedInShardsCount() {
-        return this.processingShards.values().stream()
-                .mapToInt(Map::size)
-                .sum();
+    public long getWorkQueuedInShardsCountAwaitingSelection() {
+        long count = this.processingShards.values().parallelStream()
+                .flatMap(x -> x.values().stream())
+                // missing pm.isBlocked(topicPartition) ?
```
astubbs, line 95:
```suggestion
```

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ShardManager.java
```diff
@@ -84,16 +89,22 @@ Object computeShardKey(ConsumerRecord<K, V> rec) {
...
+                .filter(WorkContainer::isAvailableToTakeAsWork)
+                .count();
+//                .mapToInt(Map::size)
+//                .sum();
```
astubbs, line 99:
```suggestion
```

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/WorkContainer.java
```diff
@@ -59,6 +62,7 @@
     /**
      * Wait this long before trying again
      */
+    // todo check it this dead code?
     private Duration retryDelay;
```
astubbs, line 66: check

parallel-consumer-vertx/src/main/java/io/confluent/parallelconsumer/vertx/VertxParallelStreamProcessor.java
```diff
@@ -80,4 +80,18 @@ void vertxHttpWebClient(BiFunction<WebClient, ConsumerRecord<K, V>, Future<HttpR
...
+     * The system will treat the messages as a set, so if an error is thrown by the user code, then all messages will be
+     * marked as failed and be retried (Note that when they are retried, there is no guarantee they will all be in the
+     * same batch again). So if you're going to process messages individually, then don't use this function.
+     * <p>
+     * Otherwise, if you're going to process messages in sub sets from this batch, it's better to instead adjust the
+     * {@link ParallelConsumerOptions#getBatchSize()} instead to the actual desired size, and process them as a whole.
+     *
+     * @see ParallelConsumerOptions#getBatchSize()
+     */
+    void batchVertxFuture(Function<List<ConsumerRecord<K, V>>, Future<?>> result);
```
astubbs, line 96: like X but...

parallel-consumer-reactor/src/main/java/io/confluent/parallelconsumer/reactor/ReactorProcessor.java
```diff
@@ -65,27 +64,48 @@ public void close(Duration timeout, DrainingMode drainMode) {
...
+     * Register a function to be applied to a batch of messages.
+     * <p>
      * Make sure that you do any work immediately in a Publisher / Flux - do not block this thread.
+     * <p>
+     * The system will treat the messages as a set, so if an error is thrown by the user code, then all messages will be
+     * marked as failed and be retried (Note that when they are retried, there is no guarantee they will all be in the
+     * same batch again). So if you're going to process messages individually, then don't use this function.
+     * <p>
+     * Otherwise, if you're going to process messages in sub sets from this batch, it's better to instead adjust the
+     * {@link ParallelConsumerOptions#getBatchSize()} instead to the actual desired size, and process them as a whole.
```
astubbs, line 76: like X but...

astubbs (commented):
getting there... lots of improvements / refactoring in this to

README.adoc
```diff
@@ -220,6 +220,7 @@ without operational burden or harming the cluster's performance
 * Solution for the https://en.wikipedia.org/wiki/Head-of-line_blocking["head of line"] blocking problem where continued failure of a single message, prevents progress for messages behind it in the queue
 * Per `key` concurrent processing, per partition and unordered message processing
 * Offsets committed correctly, in order, of only processed messages, regardless of concurrency level or retries
+* Batch version fo the API to process batches of messages in parallel instead of single messages.
```
astubbs, line 223: investigate build speed change

astubbs: might be due to the slower commit rate - 5s vs 100ms - a lot of sets "wait for one cycle", which when there's no work might be the full 5 seconds. Plus on CI the tests are sequential.

astubbs:
Target should be 100% now. Protect for negative delta.

parallel-consumer-core/src/test/java/io/confluent/parallelconsumer/AbstractParallelEoSStreamProcessorTestBase.java
```diff
@@ -205,8 +190,8 @@ protected void setupParallelConsumerInstance(ParallelConsumerOptions parallelCon
 
         subscribeParallelConsumerAndMockConsumerTo(INPUT_TOPIC);
 
-        parentParallelConsumer.setLongPollTimeout(ofMillis(DEFAULT_BROKER_POLL_FREQUENCY_MS));
-        parentParallelConsumer.setTimeBetweenCommits(ofMillis(DEFAULT_COMMIT_INTERVAL_MAX_MS));
+        parentParallelConsumer.setLongPollTimeout(ofMillis(400));
```
astubbs, line 193: if these are going back in - use the fields too

README.adoc
astubbs: deferring to
- #226

PR #43, opened 2020-12-01
Compile against released slf4j-version (1.7.30)

JorgenRingen:
`MessageFormatter#basicArrayFormat` does not exist the latest released version of slf4j-api (1.7.30). I suggest compiling against a released version of slf4j-api for better client compatibility. Kafka-clients for example compiles against 1.7.30.

Noticed this when declaring `pollAndProduce` multiple times on `ParallelStreamProcessor`. Should log an error, but instead throws "method does not exist" and crashes.


parallel-consumer-core/src/main/java/io/confluent/csid/utils/StringUtils.java
```diff
@@ -9,7 +9,7 @@
 public class StringUtils {
 
     public static String msg(String s, Object... args) {
-        return MessageFormatter.basicArrayFormat(s, args);
+        return MessageFormatter.arrayFormat(s, args).getMessage();
```
JorgenRingen, line 12: @astubbs : Was it a conscious choice to use `basicArrayFormat`? That method expects that a throwable is _not_ the last element in the array. If so, `arrayFormat(s, args, null).getMessage()` is the equivalent in 1.7.30 as far as I can see (same as `basicArrayFormat` does internally)

```
    /**
     * Assumes that argArray only contains arguments with no throwable as last element.
     */
    final public static String basicArrayFormat(final String messagePattern, final Object[] argArray) {
    	FormattingTuple ft = arrayFormat(messagePattern, argArray, null);
    	return ft.getMessage();
    }
```

JorgenRingen (commented):
Should maybe consider downgrading `logback-classic:1.3.0-alpha5` which depends on `slf4j-api:2.0.0-alpha1` to align those? Probably won't be a problem as it's test-scoped and would only cause internal test-issues.

astubbs:
Can you rebase this? we'll get it into 0.3

parallel-consumer-core/src/main/java/io/confluent/csid/utils/StringUtils.java
astubbs: cool, yeah let's get rid of the pre-release dependencies.

astubbs:
And go ahead and downgrade logback as well

astubbs (approved):

PR #51, opened 2020-12-11
Final improvements before releasing 3.0

### LINKED ISSUE #52: Bug in run length encoder - doesn't work with offset gap in source partition
For example, in compacted topics the offsets are not strictly sequential - there can be gaps. This was over looked with the RunLength encoder.

astubbs:
Fixes #52

### DIFF
--- a/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumerOptions.java
+++ b/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumerOptions.java
@@ -4,10 +4,12 @@ package io.confluent.parallelconsumer;
  * Copyright (C) 2020 Confluent, Inc.
  */
 
+import io.confluent.csid.utils.StringUtils;
 import lombok.Builder;
 import lombok.Getter;
 import lombok.ToString;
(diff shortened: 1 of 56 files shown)


parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumerOptions.java
```diff
@@ -67,34 +69,42 @@
...
+         * #PERIODIC_TRANSACTIONAL_PRODUCER}. Slower but potentially less duplicates than {@link
+         * #PERIODIC_CONSUMER_ASYNCHRONOUS} upon replay.
          */
-        CONSUMER_SYNC,
+        PERIODIC_CONSUMER_SYNC,
+
         /**
-         * Fastest option, under normal conditions will have few of no duplicates. Under failure revocery may have more
-         * duplicates than {@link #CONSUMER_SYNC}.
+         * Periodically commits offsets asynchronously. The fastest option, under normal conditions will have few of no duplicates. Under failure recovery may have more
```
JorgenRingen, line 90: "of no" -> "or no"

PR #61, opened 2020-12-20
Fix: DrainingCloseable#close should drain by default (didn't drain)

JorgenRingen:


JorgenRingen:
Hmm, maybe it's actually the doc that's wrong? ParallelEosStreamProcessor don't drain by default 🤔
```
    /**
     * Close the system, without draining.
     *
     * @see State#draining
     */
    @Override
    public void close() {
        // use a longer timeout, to cover for evey other step using the default
        Duration timeout = DrainingCloseable.DEFAULT_TIMEOUT.multipliedBy(2);
        closeDontDrainFirst(timeout);
    }
    ```

astubbs:
ah - it's a javadoc issue. sort of. the implementation overrides the inherited system. it probably shouldn't specify in the interface what it should do. fxied now.

PR #69, opened 2021-01-15
Fix for Bug 62 skip record on restart - off by one one?

### LINKED ISSUE #62: Not consuming the first message on the topic
Ok, this is a __really strange__ bug: I'm trying to use parallel-consumer with Clojure. I'm using a local Kafka cluster (only one broker) to test things on my machine.

If I send a single message, and fire up the consumer, it consumes that message. So far, so good.
If I immediately stop the consumer, send another message, and fire up the broker... nothing happens. The message is not consumed at all.

BUT, if I stop the consumer, send ANOTHER message, and fire up the consumer again, it consumes ONLY the new message... not the old one.

If I send a batch of messages, the same problem happens: it ignores the first message, and consumes the rest. Here's a video for reference: 

https://user-images.githubusercontent.com/138037/103380836-cd1bd500-4ac8-11eb-9b94-00092befe9ca.mp4

The code that causes this error is the following: please notice that there's nothing special about it - just instantiates a single consumer that prints the received message:

```clojure
(defn -main  [& args]
  (let [consumer (KafkaConsumer. (doto (java.util.Properties.)
                                       (.put "group.id" "example-1")
                                       (.put "bootstrap.servers" "localhost:9092")
                                       (.put "auto.offset.reset" "earliest")
                                       (.put "enable.auto.commit" false)
                                       (.put "key.deserializer" "org.apache.kafka.common.serialization.StringDeserializer")
                                       (.put "value.deserializer" "org.apache.kafka.common.serialization.StringDeserializer")))
        options (.. (ParallelConsumerOptions/builder)
                    (ordering ParallelConsumerOptions$ProcessingOrder/UNORDERED)
                    (defaultMessageRetryDelay (java.time.Duration/ofMillis 500))
                    (maxConcurrency 1000)
                    (commitMode ParallelConsumerOptions$CommitMode/PERIODIC_CONSUMER_ASYNCHRONOUS)
                    (consumer consumer)
                    build)
        poller (doto (ParallelStreamProcessor/createEosStreamProcessor options)
                     (.subscribe ["example"]))]
    (.poll poller (as-consumer (fn [record]
                                 (locking poller
                                   (prn :HANDLING record)))))
    (Thread/sleep 3000)
    (.close poller)))
```

astubbs:
Fixes #62 - the off by one issue would cause first message to be skipped in some situations

### DIFF
--- a/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/OffsetMapCodecManager.java
+++ b/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/OffsetMapCodecManager.java
@@ -114,10 +114,10 @@ public class OffsetMapCodecManager<K, V> {
 
         committed.forEach((tp, offsetAndMeta) -> {
             if (offsetAndMeta != null) {
-                long offset = offsetAndMeta.offset();
+                long nextExpectedOffset = offsetAndMeta.offset();
                 String metadata = offsetAndMeta.metadata();
                 try {
-                    loadOffsetMetadataPayload(offset, tp, metadata);
+                    loadOffsetMetadataPayload(nextExpectedOffset, tp, metadata);
                 } catch (OffsetDecodingError offsetDecodingError) {
                     log.error("Error decoding offsets from assigned partition, dropping offset map (will replay previously completed messages - partition: {}, data: {})",
                             tp, offsetAndMeta, offsetDecodingError);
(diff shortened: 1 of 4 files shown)


astubbs:
Released. https://repo1.maven.org/maven2/io/confluent/parallelconsumer/parallel-consumer-parent/0.3.0.1/

PR #70, opened 2021-01-15
GH Workflows build

astubbs:


astubbs:
This PR fixes the GH workflow. Atm, only compile is enabled as there seems to be race conditions with the tests when run on GH servers. I suspect they are under resourced compared to Confluent's Jenkins CI environment, which exasperates some otherwise not encountered race conditions with the test assertions which don't appear on local environments either.

PR #76, opened 2021-02-10
Reduce log-level if no results are returned from user-function

JorgenRingen:
Have some cases where it's natural to not produce any records, so reducing the log-level for this situation.


astubbs:
@JorgenRingen is that when you are conditionally producing result records? So sometimes for this function there are output records, other times there aren't? Otherwise I'm wondering why not use the void return type?

JorgenRingen:
@astubbs : Exactly, in some cases the list will be empty.

astubbs (approved):
LGTM, sorry for the wait!

PR #81, opened 2021-03-01
Parallel join technique documentation

astubbs:
Rendered docs: https://github.com/astubbs/parallel-consumer/tree/parallel-join-technique#parallel-joins

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -496,6 +496,22 @@ Use your Streams app to process your data first, then send anything needed to be
 .Example usage with Kafka Streams
 image::https://lucid.app/publicSegments/view/43f2740c-2a7f-4b7f-909e-434a5bbe3fbf/image.png[Kafka Streams Usage, align="center"]
 
+==== Parallel Joins
+
+PC can also be used to implement a parallel join against a data source, namely a Kafka Streams (KS) state store.
+
+KS state stores can be read by external threads (external to the toplogy context), and so they can be reference in the processing function of PC.
+If no item is present in the state store for the key, then it's a miss.
+If there is, you can join the data any way you please, and potentially call an external system with the joined data, or output it back to a Kafka topic using the `#pollAndProdce` API if trying to avoid https://thorben-janssen.com/dual-writes/[dual write] scnearios.
+
+image::https://lucid.app/publicSegments/view/d144c027-653c-4e77-bfa4-8ecaadba1385/image.png[]
+
+WARNING::Although using a `GlobalKTable` is not strictly nesescary, the https://docs.confluent.io/platform/current/streams/developer-guide/interactive-queries.html[extra network hop] caused by data not being colocated when using a sharded `KTable` may negate the performance benefits in some scenarios.
+For this reason, a `GlobalKTable` is recommended.
+
+WARNING::Performing a join outsode of KS relinquishes ordering efforts KS applies to populating each side of the join.
+Be careful using this technique if your operation is sensitive to the order in which data is populated in the state store vs arriving on the event stream.
+
 .Preprocess in Kafka Streams, then process concurrently
 [source,java,indent=0]
 ----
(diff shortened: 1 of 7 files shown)


README.adoc
```diff
@@ -535,6 +538,51 @@ image::https://lucid.app/publicSegments/view/43f2740c-2a7f-4b7f-909e-434a5bbe3fb
...
+----
+<1> If no matching item is present in the state store for the key, then it's a miss.
+<2> If lookup is not null, join the data any way you please, and potentially call an external system with the joined data, or output it back to a Kafka topic using the `#pollAndProdce` API if trying to avoid https://thorben-janssen.com/dual-writes/[dual write] scenarios.
+
+See the link:{project_root}/parallel-consumer-examples/parallel-consumer-example-streams/src/main/java/io/confluent/parallelconsumer/examples/streams/ParallelJoin.java[Parallel join KS example] class.
+
+WARNING: Performing a join outside of KS relinquishes ordering efforts KS applies to populating each side of the join - i.e. there is no effort to apply any ordering to the corresponding sides of this join.
+When joining within KS, this is taken care of for you.
+Be careful using this technique if your operation is sensitive to the order in which data is populated in the state store vs arriving from the event stream.
+
```
JorgenRingen, line 584: Time synchronization of joins in KS isn't done for GlobalKTables (and interactive queries?) anyway 🤔
```
 Another important difference between a KTable and a GlobalKTable is time synchronization: while processing KTable records is time synchronized based on record timestamps to all other streams, a GlobalKTable is not time synchronized.
```
https://www.confluent.io/blog/crossing-streams-joins-apache-kafka/

Do you know if there is any way to make sure that a GlobalKTable has been populated before parallel-consumer processing starts to avoid/minimize chance of join-misses? GlobalKTables are populated in a separate thread. It's usually super-fast, but have had some issues with join-misses in KS for stream-globaltable joins during startup if the globaltable is very big during startup.

src/docs/README.adoc
```diff
@@ -457,6 +460,32 @@ include::{project_root}/parallel-consumer-examples/parallel-consumer-example-str
...
+----
+<1> If no matching item is present in the state store for the key, then it's a miss.
+<2> If lookup is not null, join the data any way you please, and potentially call an external system with the joined data, or output it back to a Kafka topic using the `#pollAndProdce` API if trying to avoid https://thorben-janssen.com/dual-writes/[dual write] scenarios.
+
+See the link:{project_root}/parallel-consumer-examples/parallel-consumer-example-streams/src/main/java/io/confluent/parallelconsumer/examples/streams/ParallelJoin.java[Parallel join KS example] class.
+
+WARNING: Performing a join outside of KS relinquishes ordering efforts KS applies to populating each side of the join - i.e. there is no effort to apply any ordering to the corresponding sides of this join.
+When joining within KS, this is taken care of for you.
+Be careful using this technique if your operation is sensitive to the order in which data is populated in the state store vs arriving from the event stream.
+
```
JorgenRingen, line 487: Same as https://github.com/confluentinc/parallel-consumer/pull/81/files#r586219788

README.adoc
astubbs: GKT work a bit different.. they "fully" hydrate on application start. As in, head offset is marked for input topic at start, then topic is loaded to reach that point, then it's just whatever hits first. So, there is some effort there - just trying to point out it's different, and should be understood.

Ah yes, trying to sync with PC separately... 🤔 That's a great question. You could do something janky like read the head message, and wait until that can be queried from the GKT... Otherwise, ideally you could hook into an event listener system to know when GTK bootstrap has finished. I guess there isn't anything for this currently. Oh actually - it might be represented in the run state of KS - that's worth looking into. Thanks for the heads up!

astubbs:
Needs some suggestions on synchronising with external KS app as @JorgenRingen says..

Issue #65, opened 2021-01-14
Enhanced retry epic


astubbs:
- [x] Basic retry options - customise message failure retry delay - currently 1 seconds, enable/disable retry
- [ ] Retry max reached skip die or dlq or skip?  #196
- [x] Retry failure callback (instead or as well as DQL) implemented in #92
- [ ]  #242 
- [ ] DLQ #310

Strongly related to https://github.com/confluentinc/parallel-consumer/issues/34 - Monitor for progress and optionally shutdown (leave consumer group), skip message or send to DLQ #34

astubbs:
BTW @ismarslomic , did you see `ParallelConsumerOptions#defaultMessageRetryDelay`? Do you think there should be more documentation around it?

I've added some docs - see https://github.com/confluentinc/parallel-consumer/pull/66/commits/917e4698a272a0764cdf6554237c6e819569db0a

astubbs:
@radoslawtwardy -> https://github.com/confluentinc/parallel-consumer/issues/82 Custom retry delay duration from supplier function #82

radoslawtwardy:
>  Custom retry delay duration from supplier function 
Possibility to define custom function will be great.

astubbs:
> >  Custom retry delay duration from supplier function 
> 
> 
> 
> Possibility to define custom function will be great.

Great! Try out https://github.com/confluentinc/parallel-consumer/compare/master...astubbs:retry-delay-generator-func linked from the issue and let me know what you think of the api amd functionality.

johnbyrnejb:
Closing Issue

Issue #82, opened 2021-03-02
Custom retry delay duration from supplier function


astubbs:
Most flexible, pluggable retry delay system - user provides a function that generates a retry delay.

astubbs:
@radoslawtwardy how does this look? : 
It's a simple change. The other option I'd been thinking about, would be for the user function to throw a "retry exception" which will contain the retry delay as a field. The motivation for that solution is it gives the user ultimate context to generate the retry delay. Where as the retry delay function option only gives the user the WorkContainer as context.

radoslawtwardy:
I'm not sure how can I use it. To write custom function i must has some retry count context, or something like it. For example i want write simple algorithm:
`2^retryNb [s]`
How i can use WorkContainer to do it? Maybe i can get some previous, effective delay from it ?

astubbs:
Yes you can retrieve the number of times the record has failed amd calculate a delay from that.

astubbs:
@radoslawtwardy there is now an example of this here: https://github.com/confluentinc/parallel-consumer#retry-delay-function

PR #92, opened 2021-03-17
Implemented #65 - Custom retry delay provider

astubbs:


astubbs:
Needs example code to demonstrate: see https://github.com/confluentinc/parallel-consumer/issues/82#issuecomment-788749729

Issue #103, opened 2021-03-30
Matrix test against multiple AK and JDK versions in GH


astubbs:

astubbs:
As we are no longer able to use GH Actions, we will setup matrix builds and public PR builds in our Jenkins cluster for eternal contributors. Internal ticket has been raised and I will close this when the work is done.

astubbs:
This was done a while ago in one of my other projects. Once #220, I will copy it across.

astubbs:
Tracked in the maintained fork as astubbs/parallel-consumer#128, which carries a summary of this
issue and its current status there.

The **Confluent-hosted version** of this project is no longer maintained - its README says so.
**Parallel Consumer itself continues** in my fork: https://github.com/astubbs/parallel-consumer

Nothing in this repository is closed or changed by this comment.

PR #121, opened 2021-06-23
Add option to specify timeout for how long to wait for blocking Producer#send

JorgenRingen:


parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumerOptions.java
```diff
@@ -131,6 +131,19 @@
...
+     * returned from the user-function. Only relevant if using one of the produce-flows and providing a
+     * {@link ParallelConsumerOptions#producer}. If the timeout occurs the record will be re-processed in the user-function.
+     *
+     * Consider aligning the value with the {@link ParallelConsumerOptions#producer}-options to avoid unnecessary re-processing
+     * and duplicates on slow {@link Producer#send} calls.
+     *
+     * @see org.apache.kafka.clients.producer.ProducerConfig#DELIVERY_TIMEOUT_MS_CONFIG
+     */
+    @Builder.Default
+    private final Duration sendTimeout = Duration.ofSeconds(2);
```
JorgenRingen, line 145: @astubbs : Should we increase the default timeout? Just kept the original value for first draft.

JorgenRingen:
A little bit unsure regarding providing commit-offset-timeout as an option: https://github.com/confluentinc/parallel-consumer/blob/master/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ConsumerOffsetCommitter.java#L148

What's the reason for why timeouts isn't retried? 🤔 
```
if (take == null)
                    throw new InternalRuntimeError(msg("Timeout waiting for commit response {} to request {}", timeout, commitRequest));
```

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumerOptions.java
astubbs: Actually I think yes increase the time out. I hit this issue myself just running the test suite in parallel - slower I guess in some ways. What would you suggest using as the default?

astubbs:
> A little bit unsure regarding providing commit-offset-timeout as an option: https://github.com/confluentinc/parallel-consumer/blob/master/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ConsumerOffsetCommitter.java#L148
> 
> What's the reason for why timeouts isn't retried? 🤔
> 
> ```
> if (take == null)
>                     throw new InternalRuntimeError(msg("Timeout waiting for commit response {} to request {}", timeout, commitRequest));
> ```
> 
> This exception would stop the control-loop.

It does get retried, but it is retried through the same system as if the user's function threw an error. It was easier to do / simpler. Down side is that the message is of course also reprocessed, so not ideal.

Regarding retrying, do you have any preference for how this should occur? Probably would have it's own PR as this PR as a simple config change is trivial to merge (sorry for delay).

astubbs:
needs rebasing

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumerOptions.java
JorgenRingen: Hmm, perhaps align with default producer timeout? 🤔 
https://docs.confluent.io/platform/current/installation/configuration/producer-configs.html#producerconfigs_delivery.timeout.ms

JorgenRingen:
> It does get retried, but it is retried through the same system as if the user's function threw an error. It was easier to do / simpler. Down side is that the message is of course also reprocessed, so not ideal.
> 
> Regarding retrying, do you have any preference for how this should occur? Probably would have it's own PR as this PR as a simple config change is trivial to merge (sorry for delay).

Perfectly understandable 👍 

Agree that reprocessing in user-function should be limited as much as possible, so consumer/producer-errors should preferably be retried in the client lib itself if possible. In this specific case it might just be a matter of aligning the timeout so that the consumer is able to retry exhaustively when. Perhaps https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html#consumerconfigs_request.timeout.ms is the one 🤔

```
Duration timeout = ParallelEoSStreamProcessor.DEFAULT_TIMEOUT;
                CommitResponse take = commitResponseQueue.poll(timeout.toMillis(), TimeUnit.MILLISECONDS); // blocks, drain until we find our response
```

NB: This exception actually caused the control-loop to stop as far as I remember, but msg was reprocessed after application restart :)

astubbs:
> NB: This exception actually caused the control-loop to stop as far as I remember, but msg was reprocessed after application restart :)

Sorry, what exception caused the system to shutdown?

astubbs:
I think you need to mark this as ready for review?

JorgenRingen:
> > NB: This exception actually caused the control-loop to stop as far as I remember, but msg was reprocessed after application restart :)
> 
> Sorry, what exception caused the system to shutdown?

If offset-commit timed out (10sec default): https://github.com/confluentinc/parallel-consumer/blob/master/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/ConsumerOffsetCommitter.java#L156

> I think you need to mark this as ready for review?
Sorry, rebased and marked as ready! Build failed, but not possible to see details: https://jenkins.confluent.io/job/confluentinc-pr/job/parallel-consumer/job/PR-121/3/display/redirect

astubbs:
Ah sorry, our build system is under maintenance atm.

> ERROR: We have disabled builds for pull requests created by the open source communitydue to build system maintenance.We are working to finish it soon.

I'll check what the eta is.

astubbs:
ok @JorgenRingen , issue has been raised internally - and we should have a public PR builder setup soon, along with matrix builds for multiple AK versions like we used to have with GH actions.

astubbs:
FYI Matrix test against multiple AK and JDK versions in GH #103

astubbs:
Ok @JorgenRingen , if you rebase now it should build

astubbs (approved):

PR #126, opened 2021-07-09
Remove static state manipulation that tests use - breaks parallel testing

astubbs:


astubbs:
Made redundant by #405

PR #133, opened 2021-07-13
Base class refactor - removes core api from extension modules

### LINKED ISSUE #99: Remove non vertx operations from vertx implementation
Will probably need to refactor out a common parent to do it nicely. Shouldn't be too hard though. 

People might instantiate a vertx implementation but mistaking luck use the vanilla methods, where the thread pool size is only one. 

I can't think right now of a good reason to keep them both available in the vertx implementation.

astubbs:
fixed #99 Completely remove non vertx operations from vertx implementation be refactoring out an abstract base class that Vert.x module can extend, instead of extending the core PC.

This is because the threading model is so different, it's totally invalid to call the core methods from the extension modules. There's also zero advantage in having them both available.


parallel-consumer-vertx/src/test/java/io/confluent/parallelconsumer/vertx/VertxNonVertxOperations.java
```diff
@@ -4,24 +4,20 @@
...
 public class VertxNonVertxOperations extends ParallelEoSStreamProcessorTest {
 
-    @Override
-    protected ParallelEoSStreamProcessor initAsyncConsumer(ParallelConsumerOptions parallelConsumerOptions) {
-        VertxOptions vertxOptions = new VertxOptions();
-        Vertx vertx = Vertx.vertx(vertxOptions);
-        parallelConsumer = new VertxParallelEoSStreamProcessor<>(vertx, WebClient.create(vertx), parallelConsumerOptions);
-        return parallelConsumer;
-    }
+    // todo delete - invalid test
```
astubbs, line 14: should delete class

PR #137, opened 2021-07-21
Make parallel consumer java ee compliant

System0X:
As far as I looked I only found two places where ExecutorService or ThreadFactory is being used.


astubbs (changes requested):
Thanks! Can you change the log level to debug and include the exception message.

astubbs:
I’m not that familiar with EE anymore - should the string be configurable with a default?

System0X:
> should the string be configurable with a default?

Yes, it should. Is there a default way you deal with it?

astubbs:
Yeah take a look in the options class.

System0X:
The changes have been made😄

astubbs (approved):

astubbs:
Thanks!

PR #139, opened 2021-07-22
ux: Adds a warning when no SLF4J loger detected

astubbs:


astubbs:
Slf4j logs it's own warning

PR #140, opened 2021-07-23
fix: #99 Don't allow use of core methods from the vert.x or reactor module

astubbs:
The threading system for the vert.x module is not compatible with the core module. Use the core module directly instead.

Partial temp fix for #99


astubbs:
Superseded by https://github.com/confluentinc/parallel-consumer/pull/133.

PR #145, opened 2021-07-27
show record topic in slow-work warning message

ghmulti:
Hi,
We are using "WARN" log level and switching to "INFO"/"DEBUG" level will lead to a lot of noise - though "Warning: {} records in the queue have been waiting longer than {}" is important enough but lacks details to troubleshoot the potential problem.
Shall at least topic names be added to that log message?


astubbs:
Let me know how you get on..

PR #197, opened 2022-02-15
feature: Retry count available in header

astubbs:
- [x] docs
- [x] update / link to from issues
- [x] include issue number
- [x] Remove WorkContainer from any public API - e.g. delay provider in options

Spirit of: 
- #65 

Redundant by 
- #216 

Blocked by:
- #223


parallel-consumer-examples/parallel-consumer-example-core/src/main/java/io/confluent/parallelconsumer/examples/core/CoreApp.java
```diff
@@ -91,7 +92,8 @@ void runPollAndProduce() {
 
         // tag::exampleProduce[]
         parallelConsumer.pollAndProduce(record -> {
-                    var result = processBrokerRecord(record);
+            int failedCount = ParallelConsumerRecord.getFailedCount(record);
```
astubbs, line 95: this was a start, need to complete an example usage and add to docs

astubbs:
Redundant by 
- #216

PR #204, opened 2022-03-01
vert.x: Run user functions on a Vert.x vertical, instead of a Java ThreadPool

astubbs:
May be overkill?


eddyv:
Closing - Stale.

Issue #201, opened 2022-02-24
Refactor: Live tracking of offsets as they change, so we don't need to scan for them


astubbs:
The state for offsets is really known at runtime, but not recorded. The scan for them in PartitionMonitor#findCompletedEligibleOffsetsAndRemove isn't really required if this is implemented.

See:
- #46

PR #218, opened 2022-03-08
refactor: Partition state tracking instead of search

astubbs:
Blocked by:
- #224

Related:
- #201


astubbs (commented):
looking good

astubbs (commented):
nearly there

PR #234, opened 2022-03-16
docs: Embed changelog in readme

astubbs:


astubbs:
Doesn't NEED level offset for it to look ok, but would be a nice to have.

PR #219, opened 2022-03-08
refactor: Queue unification

astubbs:
Massive simplification, much improved semantics around poller notifying of work incoming. 

Also address:
- #236

PR #237, opened 2022-03-17
fixes #236: Shard starvation under some circumstances

### LINKED ISSUE #236: Under some conditions, a shard (by partition or key), can get starved for attention
The simplest scenario of this issue, is:

using PARTITION ordering
max concurrency set to 2

1. consumer is able poll 10 messages, the first 5 for partition 0, the next 5 for partition 1
2. when PC requests work from work manager, it first requests 2 work units
3. work manager then ingests 2 units from it's queue, but to shard partition 0
4. then shard iteration begins, partition order restriction is set, so only takes a single entry from the ONLY partition shard of zero
5. so result is only a single record is returned, even through there are 5 in the queue for the 2nd partition that aren't "realised"
6. on next request to WM, 1 more unit is requested to hit concurrency target - and this is also ingested from the 3 remaining for partition 0, and the rest for partition 1 are still undiscovered, and so on and so on - extrapolate out...

solution:
ingest the entire work queue into shards every round, which is effectively also what #219

Also fixed by upcoming RR:
- #219

astubbs:
see:
- #236

made redundant by:
- #219

### DIFF
--- a/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/offsets/OffsetSimultaneousEncoder.java
+++ b/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/offsets/OffsetSimultaneousEncoder.java
@@ -61,8 +61,7 @@ public class OffsetSimultaneousEncoder {
      * @see #packSmallest()
      */
     @Getter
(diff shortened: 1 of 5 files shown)


astubbs:
Redundantly fixed by:
-  #219

PR #303, opened 2022-05-13
Experiment: extend Consumer and Function

astubbs:
Experiment: extend Consumer and Function - centralises documentation and makes usage more explicit

### Checklist

- [ ] Changelog


rkolesnev:
LGTM - nice addition i think.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/RecordProcessor.java
```diff
@@ -0,0 +1,48 @@
...
+
+        /**
+         * Like {@link PollConsumer#accept(PollContext)} but also returns records to be produced back to Kafka.
+         *
+         * @param records the Kafka records to process
+         * @return the function result
+         * @see PollConsumer#accept(PollContext)
+         */
+        @Override
+        List<ProducerRecord<K, V>> apply(PollContext records);
```
rkolesnev, line 45:
```suggestion
        List<ProducerRecord<K, V>> apply(PollContext<K,V> records);
```

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/RecordProcessor.java
```diff
@@ -0,0 +1,48 @@
...
+         * @param records the Kafka records to process
+         * @see PCRetriableException
+         * @see ParallelConsumerOptions#getRetryDelayProvider()
+         * @see ParallelConsumerOptions#getDefaultMessageRetryDelay()
+         */
+        void accept(PollContext<K, V> records);
+    }
+
+    @FunctionalInterface
+    interface PollConsumerAndProducer<K, V> extends java.util.function.Function<PollContext<K, V>, List<ProducerRecord<K, V>>> {
```
rkolesnev, line 35: Staying with PollConsumerAndProducer ? Would MessageTransformer or Processor or Mapper sound better given the messages in / out nature?

parallel-consumer-examples/parallel-consumer-example-core/src/main/java/io/confluent/parallelconsumer/examples/core/CoreApp.java
```diff
@@ -141,6 +141,9 @@ void maxRetries() {
         final int maxRetries = 10;
         final Map<ConsumerRecord<String, String>, Long> retriesCount = new ConcurrentHashMap<>();
 
+        pc.pollAndProduceMany();
+        pc.poll(records ->);
+
```
rkolesnev, line 146: presumably - to be removed?

rkolesnev (changes requested):

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/RecordProcessor.java
astubbs: yeah - other thing is, if it affects your ideas - is it can produce many messages, not just one, from a single input

parallel-consumer-examples/parallel-consumer-example-core/src/main/java/io/confluent/parallelconsumer/examples/core/CoreApp.java
astubbs: yeah I was playing with what the api ended up looking like

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/RecordProcessor.java
rkolesnev: Would it be worth going to KStream DSL naming? have a map and flatmap where map takes 1 in - 1 out, and flatmap allows 1->many ? The typing needs fixed to allow transform of key / value anyway - so maybe can be done in separate PR that deals with it. Or Transform (with or without flatTransform).
Thinking about it - "Transformer" sounds best to me. 
Are side effects allowed / expected in this scenario (i.e. external system calls for processing / enrichment)? If so something neutral like Processor would be ok i think.

PR #307, opened 2022-05-17
fix #306: Change Truth-Generator dependency from compile to test

### LINKED ISSUE #306: Can’t use PC without adding package cloud to our maven repo proxy
This is because one dependency - truth generator api - is published only to package cloud. Publishing TG to repo1 (maven central) would avoid this.

Building and running PC was tested in a clean environment on Confluent’s build system, so the package cloud repo works, but in some environments, it’s not able to be reached automatically without extra config steps to the proxy setup. 

https://github.com/confluentinc/parallel-consumer/issues/264#issuecomment-1128532160

Workaround: configure maven proxy to allow access to package cloud. 

edit:
Ah - I just realised it was incorrectly set as a compile time dependency, which is incorrect. It's only needed to run the tests. Will fix in:
- #306 
and release in 0.5.1.0

astubbs:
fixes:
- #306

TG was incorrectly marked as compile dependency, when it should be test.

This was preventing use of the library if access tp package cloud's repo wasn't possible.

Access atm to package cloud is still required to build run the tests on the project however.


astubbs (commented):
lgtm

Issue #314, opened 2022-05-24
With KEY ordering, option to combine queues from different partitions or topics


astubbs:
In key mode, if records of a given key exist in
- multiple partitions of a topic (and those partitions are assigned to me)
- or when subscribing to multiple topics which have matching keys, have an option to have these either combined into a single queue, or kept separate. 

If combined, option to choose how records are ordered relatively - e.g. via time stamp extractor, offset or publish time stamp. 

Related:
- #184

phanivankadari:
We are looking at similar use case where we have multiple topics (for prioritization and fairness usecases) and we wanted key grouping across all topics related to a microservice (wildcard topic or list of topics specified). I suppose the current key grouping is supported to work for messages for a given key at a single partition level - is this assumption correct?

I see the PR is closed as stale but did not have much comments if had any other issues. Any updates on the feasibility of this feature request?

JorgenRingen:
Any plans to implement this i the near future? :-)

rkolesnev:
Not really - it would be quite niche feature with a lot of caveats - main one - there is really be no easy and bulletproof way to guarantee deterministic ordering of messages with same key when consumed from different topics across rebalances, subscription changes - even poll batching / sizing may play a role in how messages are read from the topics and ultimately sequenced.

astubbs:
Tracked in the maintained fork as astubbs/parallel-consumer#150, which carries a summary of this
issue and its current status there.

The **Confluent-hosted version** of this project is no longer maintained - its README says so.
**Parallel Consumer itself continues** in my fork: https://github.com/astubbs/parallel-consumer

Nothing in this repository is closed or changed by this comment.

PR #315, opened 2022-05-25
fixes #184: Fix multi topic subscription with KEY order by adding topic to shard key

### LINKED ISSUE #184: Subscribing to two or more topics with KEY ordering, results in messages of the same Key never being processed
Steps to reproduce:

1) Started zoo keeper and a single broker on my MacBook
2) Create three topics each with one partition and a replication factor of one. The topic names differ only in the last character
3) Using a  producer program pump 333 messages into each topic with almost no delay
4) Start up  Java program that uses the parallel consumer to subscribe to and consume the messages from the three topics using KEY ordering.
5) Notice that not all of the messages show up in the polling lambda function and the Kafka topic details show an expected CURRENT-OFFSET for only one of the topics.

If one restarts the java program then additional (but not all) messages are consumed. Restart a couple more times and then all the messages have been consumed the topic details show the correct CURRENT-OFFSET for each topic.

If one deletes/purges the topics and pumps 333 messages into the first topic, then 333 more messages into the first and second topic then 333 each of the three topics then everything works fine. 

I enabled trace logging for a subsequent run and see 

Returning 3 fetched records at offset FetchPosition{offset=232, offsetEpoch=Optional.empty, currentLeader=LeaderAndEpoch{leader=Optional[192.168.2.58:9092 (id: 0 rack: null)], epoch=0}} for assigned partition myTopicprefix-topic-1-0",

But later in the logs only see one
{"timestamp":"2022-01-05T16:45:43.143Z","message":"asyncPoll - Consumed a record (232)
And
{"timestamp":"2022-01-05T16:45:47.999Z","message":"asyncPoll - user function finished ok.

If the above is performed with just one topic there is no problem. If the above is done with two fresh topics the issue also occurs.

If UNORDERED is used instead of KEY then everything works fine.

astubbs:
- [x] Changelog

Closes: #184, #318
- #184
- #318

Blocked by:
- [x] https://github.com/astubbs/truth-generator/pull/117
- [x] #332 

Related:
- #318
- #314


parallel-consumer-core/src/test/java/io/confluent/parallelconsumer/truth/ConsumerSubject.java
```diff
@@ -0,0 +1,84 @@
...
+        return map.values().stream()
+                .findFirst()
+                .orElse(
+                        check("getCommitHistory(%s)", topic.name())
+                                .about(commitHistories())
+                                .that(new CommitHistory(List.of())));
+    }
+
+    public Map<TopicPartition, CommitHistorySubject> hasCommittedToPartition(Set<NewTopic> topic) {
+        Set<TopicPartition> partitions = topic.stream().map(newTopic -> new TopicPartition(newTopic.name(), 0)).collect(Collectors.toSet());
```
astubbs, line 65: Rename method to work not with topics but TopicPartitions instead of default partition number

PR #353, opened 2022-07-15
fixes #185: Broker status informer

### LINKED ISSUE #185: Expose way to know if connection to Broker is lost then later regained
A client may have the need to expose its health to the rest of the system (Kubernetes) and so client needs a way to know if the parallel consumer connection to the broker has been lost. It later needs to be informed if the connection has been re-established. 

Related:
- #27 
- #345 
- #353

astubbs:
Description...

### Checklist

- [ ] Changelog


fowlerp-qlik:
> @fowlerp-qlik how does this interface look?

LGTM :)

astubbs:
As per #185, I don't see a way to be informed from the consumer that a connection has been lost.

PR #376, opened 2022-08-10
fixes #375: Mockito dependency should be test scoped

### LINKED ISSUE #375: Mockito dependency needs to be test scoped
From parent pom.xml, mockito dependency is not scoped as test.
https://github.com/confluentinc/parallel-consumer/blob/master/pom.xml#L335-L339
```
        <dependency>
            <groupId>org.mockito</groupId>
            <artifactId>mockito-junit-jupiter</artifactId>
            <version>${mockito.version}</version>
<!--            <scope>test</scope> <<<<<<<<<<<< MISSING --> 
        </dependency>
```

wonmo01:
From parent pom.xml, mockito dependency is not scoped as test.
https://github.com/confluentinc/parallel-consumer/issues/375

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -336,6 +336,7 @@
             <groupId>org.mockito</groupId>
             <artifactId>mockito-junit-jupiter</artifactId>
             <version>${mockito.version}</version>
+            <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


astubbs:
Why did you close this?

astubbs (approved):
lgtm

PR #431, opened 2022-10-03
[DP-9090]add run_semaphore_public

lyao-77:
A new parameter "run_semaphore_public" is added into manifest and is used to whitelist project that are public on github but could run semaphore plugin.
For new project, if we want to avoid this kind of error and change the manifest manually, one possible solution is to add a new parameter of "run_semaphore_public" in Jenkins so that this parameter will be passed in to projects.json, which will add this parameter to manifest by default.

### DIFF
--- a/service.yml
+++ b/service.yml
@@ -10,3 +10,4 @@ github:
     repo_name: confluentinc/parallel-consumer
 semaphore:
     enable: true
+    run_semaphore_public: true


service.yml
```diff
@@ -10,3 +10,4 @@ github:
     repo_name: confluentinc/parallel-consumer
 semaphore:
     enable: true
+    run_semaphore_public: true
```
acourtneybrown, line 13: i don't think this is the right place to allow this configuration.  it might make sense in `projects.json`, but since the team owning this repo can approve/merge a PR adding this config, it wouldn't really provide any roadblock.

acourtneybrown (changes requested):

PR #436, opened 2022-10-12
refractor: Move PSM stuff to PS, unify PS collections, change Set to List

astubbs:
- Very big improvement to cohesiveness
- Simpler, more meaningful, unification of collections in PartitionState
- Better API contracts


parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -562,30 +561,6 @@ private boolean isResponsibleForCommits() {
...
-     * Block the calling thread until no more messages are being processed.
-     * <p>
-     * Used for testing.
-     *
-     * @deprecated no longer used, will be removed in next version
-     */
-    // TODO delete
-    @Deprecated
-    @SneakyThrows
-    public void waitForProcessedNotCommitted(Duration timeout) {
```
astubbs, line 575: double check

astubbs (commented):
last one

PR #439, opened 2022-10-12
fix: Improvements to encoding ranges (int vs long)

astubbs:
Replace integer offset references with long - use Long everywhere we deal with offsets, and where we truncate down, do it exactly, detect and handle truncation issues.

### Checklist

- [x] Changelog


astubbs (commented):
lgtm

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/offsets/OffsetSimultaneousEncoder.java
```diff
@@ -104,9 +107,9 @@ public OffsetSimultaneousEncoder(long baseOffsetToCommit, long highestSucceededO
         }
 
         // BitSet only support Integer.MAX_VALUE bits
-        length = (int) bitsetLengthL;
+        lengthBetweenBaseAndHighOffset = (int) bitsetLengthL;
```
astubbs, line 110: check

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/offsets/BitSetEncoder.java
```diff
@@ -46,12 +46,15 @@ public class BitSetEncoder extends OffsetEncoder {
...
+     */
     public static final Integer MAX_LENGTH_ENCODABLE = Integer.MAX_VALUE;
 
     @Getter
     private final BitSet bitSet;
 
-    private final int originalLength;
+    private final long originalLength;
 
     private Optional<byte[]> encodedBytes = Optional.empty();
```
rkolesnev, line 60: doesn't this one needs to change to long as well?
`    public BitSetEncoder(int length, OffsetSimultaneousEncoder offsetSimultaneousEncoder) throws BitSetEncodingNotSupportedException {
`

rkolesnev (approved):
One comment - BitSet constructor still accepts Integer.
otherwise LGTM.

astubbs (commented):
lgtm

parallel-consumer-core/src/main/java/io/confluent/csid/utils/Range.java
```diff
@@ -57,23 +76,17 @@ public void remove() {
...
     }
 
-    public List<Integer> list() {
-        ArrayList<Integer> integers = new ArrayList<Integer>();
-        forEach(integers::add);
+    /**
+     * Potentially slow, but useful for tests
+     */
+    public List<Integer> listAsIntegers() {
+        ArrayList<Integer> integers = new ArrayList<>();
```
rkolesnev, line 83:
```suggestion
        List<Integer> integers = new ArrayList<>();
```
Just catches eye at a glance...

parallel-consumer-core/src/main/java/io/confluent/csid/utils/Range.java
```diff
@@ -57,23 +76,17 @@ public void remove() {
...
+        forEach(e -> integers.add(Math.toIntExact(e)));
```
rkolesnev, line 84: wouldnt that be slow? or its just to avoid cast warning?
can just test 
```
if(limit<Integer.MAX_VALUE){
     forEach(e -> integers.add((int)e));
}
```
or alternatively - integers.addAll(IntStream.range(0,limit).Collectors.toList()); -- if need to ensure ArrayList / i.e. list with Add support etc...
or just IntStream.range(0,limit).Collectors.toList() if dont care if can add to it later...

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/offsets/BitSetEncoder.java
```diff
@@ -77,7 +74,7 @@ public BitSetEncoder(long length, OffsetSimultaneousEncoder offsetSimultaneousEn
 
         this.originalLength = length;
 
-        // prep bit set buffer
+        // prep bit set buffer, range check above
         bitSet = new BitSet(Math.toIntExact(length));
```
rkolesnev, line 78: Bike shedding alert :) 
just generally thinking - would it be worth doing try / catch ArithmeticException instead of explicit check for length above? atm - you are relying on the separate check above - which is fine as such - but at same it pure sequencing reliance.

astubbs (commented):
lgtm

astubbs (commented):
lgtm

Issue #50, opened 2020-12-11
Feature: When subscribing to multiple topics, have the ability to priorities topics over others


astubbs:
When subscribing to multiple topics, it may be useful to prioritise processing of some topics over others.

To avoid starving low priority topics, we could use a system where we only _reserve_ a portion of our processing to the higher priority topics. 

For example (high priority means more important):

Topic A -> priority 3
Topic B -> priority 2
Topic C -> priority 1
Topic D -> priority 1

90% of concurrency is available to priority 3 messages _if_ and _only if_ it's needed (enough messages are available for processing.
75% is available for priority 2.

Percentages could be derived or explicitly provided.

As with this system, we can only prioritise messages we've downloaded (but not necessarily processed into our queues), it might be useful to allow uses to explicitly tell the system to keep a larger buffer downloaded than what is normally necessary to satisfy concurrent processing demand.

An extension of this, we could experiment with prioritising messages _within_ a topic as well.

johnbyrnejb:
Closing Issue

PR #441, opened 2022-10-18
feature #50: Features/queue priority

astubbs:
Implements #50 
- #50

### Checklist

- [ ] Changelog


eddyv:
Closing - Stale.

C0mbatwombat:
This would be really improve a notification system use-case. Can imagine implementation is not simple, but if there is one Kafka construct where it would make sense supporting prioritization, it is the parallel-consumer.

PR #443, opened 2022-10-19
Python Support

Robbie-Palmer:
Add a Python package for Parallel Consumer that provides a wrapper for the API

TODO:
- Publish python package to pypi
- Speed up startup time by checking for existing JRE

### Checklist

- [ ] Documentation
    - [X] Module level
    - [ ] Root level
- [ ] Changelog

### DIFF
--- a/parallel-consumer-python/pyallel_consumer/pom.xml
+++ b/parallel-consumer-python/pyallel_consumer/pom.xml
@@ -1,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd"
+         xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>io.confluent.parallelconsumer</groupId>
+    <artifactId>python</artifactId>
+    <version>0.1.0.SNAPSHOT</version>
+    <properties>
+        <jackson.version>2.13.2</jackson.version>
+    </properties>
+    <dependencies>
+        <dependency>
+            <groupId>io.confluent.parallelconsumer</groupId>
+            <artifactId>parallel-consumer-core</artifactId>
+            <version>0.5.2.3</version>
+        </dependency>
+        <dependency>
+            <groupId>io.confluent</groupId>
+            <artifactId>kafka-avro-serializer</artifactId>
+            <version>5.3.0</version>
+        </dependency>
+        <dependency>
+            <groupId>io.confluent</groupId>
+            <artifactId>kafka-json-serializer</artifactId>
+            <version>7.2.1</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>2.13.2.2</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+            <version>${jackson.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.codehaus.jackson</groupId>
+            <artifactId>jackson-mapper-asl</artifactId>
+            <version>1.9.13</version>
+        </dependency>
+    </dependencies>
+</project>
--- a/.mvn/wrapper/maven-wrapper.properties
+++ b/.mvn/wrapper/maven-wrapper.properties
@@ -1,0 +1,18 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.8.6/apache-maven-3.8.6-bin.zip
+wrapperUrl=https://repo.maven.apache.org/maven2/org/apache/maven/wrapper/maven-wrapper/3.1.1/maven-wrapper-3.1.1.jar
--- a/mvnw
+++ b/mvnw
@@ -1,0 +1,287 @@
+#!/bin/sh
+# ----------------------------------------------------------------------------
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+# ----------------------------------------------------------------------------
+
+# ----------------------------------------------------------------------------
+# Apache Maven Wrapper startup batch script, version 3.1.1
+#
+# Required ENV vars:
+# ------------------
+#   JAVA_HOME - location of a JDK home dir
+#
+# Optional ENV vars
+# -----------------
+#   MAVEN_OPTS - parameters passed to the Java VM when running Maven
+#     e.g. to debug Maven itself, use
+#       set MAVEN_OPTS=-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=8000
+#   MAVEN_SKIP_RC - flag to disable loading of mavenrc files
+# ----------------------------------------------------------------------------
+
+if [ -z "$MAVEN_SKIP_RC" ] ; then
+
+  if [ -f /usr/local/etc/mavenrc ] ; then
+    . /usr/local/etc/mavenrc
+  fi
+
+  if [ -f /etc/mavenrc ] ; then
+    . /etc/mavenrc
+  fi
+
+  if [ -f "$HOME/.mavenrc" ] ; then
+    . "$HOME/.mavenrc"
+  fi
+
+fi
+
+# OS specific support.  $var _must_ be set to either true or false.
+cygwin=false;
+darwin=false;
+mingw=false
+case "`uname`" in
+  CYGWIN*) cygwin=true ;;
+  MINGW*) mingw=true;;
+  Darwin*) darwin=true
+    # Use /usr/libexec/java_home if available, otherwise fall back to /Library/Java/Home
+    # See https://developer.apple.com/library/mac/qa/qa1170/_index.html
+    if [ -z "$JAVA_HOME" ]; then
+      if [ -x "/usr/libexec/java_home" ]; then
+        JAVA_HOME="`/usr/libexec/java_home`"; export JAVA_HOME
+      else
+        JAVA_HOME="/Library/Java/Home"; export JAVA_HOME
+      fi
+    fi
+    ;;
+esac
+
+if [ -z "$JAVA_HOME" ] ; then
+  if [ -r /etc/gentoo-release ] ; then
+    JAVA_HOME=`java-config --jre-home`
+  fi
+fi
+
+# For Cygwin, ensure paths are in UNIX format before anything is touched
+if $cygwin ; then
+  [ -n "$JAVA_HOME" ] &&
+    JAVA_HOME=`cygpath --unix "$JAVA_HOME"`
+  [ -n "$CLASSPATH" ] &&
+    CLASSPATH=`cygpath --path --unix "$CLASSPATH"`
+fi
+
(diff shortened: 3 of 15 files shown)


parallel-consumer-python/.gitignore
```diff
@@ -0,0 +1,129 @@
+# Byte-compiled / optimized / DLL files
```
astubbs, line 1: good lord that's a long list

src/docs/README_TEMPLATE.adoc
```diff
@@ -1,7 +1,7 @@
 //
 // STOP!!! Make sure you're editing the TEMPLATE version of the README, in /src/docs/README_TEMPLATE.adoc
 //
-// Do NOT edit /README_TEMPLATE.adoc as your changes will be overwritten when the template is rendered again during
+// Do NOT edit /README.adoc as your changes will be overwritten when the template is rendered again during
```
astubbs, line 4: oh! thanks :)

src/docs/README_TEMPLATE.adoc
```diff
@@ -556,6 +556,37 @@ From the `Clients` view, get the connection information customized to your clust
 .. Use these settings presented to https://docs.confluent.io/clients-kafka-java/current/overview.html[configure your clients].
 . Use these clients for steps outlined in the <<common_preparation>> section.
 
+
+[[python]]
+=== Python
```
astubbs, line 561: also add to feature list?

src/docs/README_TEMPLATE.adoc
```diff
@@ -556,6 +556,37 @@ From the `Clients` view, get the connection information customized to your clust
...
+...
+options = ParallelConsumerOptions.builder() \
+        .ordering(ParallelConsumerOptions.ProcessingOrder.KEY) \
+        .maxConcurrency(1000) \
+        .consumer(consumer) \
+        .producer(producer) \
+        .build()
+processor = ParallelStreamProcessor.createEosStreamProcessor(options)
+processor.subscribe([topic])
+processor.poll(lambda o: print(f"Concurrently processing a record: {o}"))
```
astubbs, line 585: <3

src/docs/README_TEMPLATE.adoc
```diff
@@ -556,6 +556,37 @@ From the `Clients` view, get the connection information customized to your clust
...
+```
+
+See the `parallel-consumer-python` [README](./parallel-consumer-python/README.adoc) for more information.
```
astubbs, line 588: we should do more of that with the rest of the docs :)

parallel-consumer-python/pyallel_consumer/pom.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd"
+         xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>io.confluent.parallelconsumer</groupId>
+    <artifactId>python</artifactId>
```
astubbs, line 7: lets stick with the rest of the convention for now
```suggestion
    <artifactId>parallel-consumer-python</artifactId>
```

parallel-consumer-python/pyallel_consumer/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+    <version>0.1.0a0</version>
```
astubbs, line 8: alpha 0?

parallel-consumer-python/pyallel_consumer/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+    <properties>
+        <jackson.version>2.13.2</jackson.version>
+    </properties>
+    <dependencies>
+        <dependency>
+            <groupId>io.confluent.parallelconsumer</groupId>
+            <artifactId>parallel-consumer-core</artifactId>
+            <version>0.5.2.3</version>
```
astubbs, line 16: even though it's not reused yet - lets extract this as a property as it's so critical..
in fact, let's use the parent' id - it'll get release as non snapshot when the rest of PC is released

parallel-consumer-python/pyallel_consumer/pom.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd"
+         xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <modelVersion>4.0.0</modelVersion>
```
astubbs, line 5: add parent like other modules?

parallel-consumer-python/pyallel_consumer/pom.xml
astubbs: so it will be managed by maven rep management

parallel-consumer-python/pyallel_consumer/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>io.confluent.parallelconsumer</groupId>
+            <artifactId>parallel-consumer-core</artifactId>
+            <version>0.5.2.3</version>
+        </dependency>
+        <dependency>
+            <groupId>io.confluent</groupId>
+            <artifactId>kafka-avro-serializer</artifactId>
+            <version>5.3.0</version>
```
astubbs, line 21: remove - managed by maven rep management

```suggestion
```

parallel-consumer-python/pyallel_consumer/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.confluent</groupId>
+            <artifactId>kafka-json-serializer</artifactId>
+            <version>7.2.1</version>
```
astubbs, line 26: remove - managed by maven rep management

```suggestion
```

parallel-consumer-python/pyallel_consumer/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>2.13.2.2</version>
```
astubbs, line 31: are these and the below, used directly?

astubbs (changes requested):
Awesome. Performance test in separate PR?

Try adding the parent pom relationship first - I'm not sure that will actually work with how it's being used. but lets try.

parallel-consumer-python/.gitignore
Robbie-Palmer: I can remove most of this, this is just the default .gitignore GitHub gives for Python

parallel-consumer-python/pyallel_consumer/pom.xml
Robbie-Palmer: Yeh I tried doing `SNAPSHOT` but PyPi didn't like it
I then tried just `a` for alpha but the packaging tools automatically added a 0 to the end if I didn't give one
So now explicitly going with `a0` which yes stands for alpha 0

parallel-consumer-python/pyallel_consumer/pom.xml
Robbie-Palmer: This is a static file used independently of the other files
It is shipped with the python package and on import of `pyallel_consumer` the JRE is installed, followed by using maven to install these dependencies by using this static pom file
See the `_Maven` class in `__init__.py`
```python
maven = _Maven()
maven.install_dependencies()
```

So I can either include the other pom files, but then maven may try to do too much?
Or I can dynamically load some of these properties in `setup.py` when creating the python package

parallel-consumer-python/pyallel_consumer/pom.xml
Robbie-Palmer: No, they're used by the Kafka client, but even then probably only because I've setup the examples to use the Kafka JSON serdes
Have included them in the alpha so it's possible to use this with JSON
But would probably need to take similar dependencies for avro/protobuf etc

I don't know how to best handle these optional dependencies
I can setup the python package to have optional dependencies, and if those settings are chosen on install I could cause these to be installed
But that then requires maintaining these optional java dependency mappings in python too
When ideally installing the kafka client would just make it usable with any format

parallel-consumer-python/pyallel_consumer/pom.xml
Robbie-Palmer: `kafka-avro-serializer` isn't included in the rest of PC

parallel-consumer-python/pyallel_consumer/pom.xml
Robbie-Palmer: This dependency isn't specified in the rest of PC

parallel-consumer-python/pyallel_consumer/pom.xml
Robbie-Palmer: If I remove them I get these errors:

Without jackson-core:
```
java.lang.java.lang.ClassNotFoundException: java.lang.ClassNotFoundException: com.fasterxml.jackson.module.paramnames.ParameterNamesModule
```
Without jackson-mapper-asl
```
java.lang.java.lang.NoClassDefFoundError: java.lang.NoClassDefFoundError: com/fasterxml/jackson/core/Versioned
```
Seems I maybe can remove jackson-databind

parallel-consumer-python/pyallel_consumer/pom.xml
Robbie-Palmer: I have it pulling from the project root pom now
But means the python package fails as when it executes `./mvnw install` it fails to find `0.5.2.4-SNAPSHOT`

astubbs:
Will look to merge this for 0.5.3

parallel-consumer-python/pyallel_consumer/pom.xml
Robbie-Palmer: Have fixed this
The parallel-consumer-core jar file is now packaged inside the python package and installed locally on startup on the client
This means the python package can now be built and used with SNAPSHOT versions of the java library, rather than just versions available on a public maven repo

parallel-consumer-python/tests/load/test_java_clients.py
```diff
@@ -0,0 +1,81 @@
...
+        sleep_a_bit()
+        count.getAndIncrement()
+
+    processor.poll(process_message)
+    success_criterion = lambda: count.get() >= num_messages
+    wait().at_most(*MAX_TIMEOUT).until(success_criterion)
+    assert success_criterion()
+
+
+def test_async_consume_and_process(loaded_topic: str, consumer: KafkaConsumer, transactional_producer: KafkaProducer,
```
Robbie-Palmer, line 67: @astubbs it would be great to have a load test which is more representative of a use case where parallel consumer shines if you have advice on how to set this up
This is a port of the Java integration test [LoadTest](https://github.com/confluentinc/parallel-consumer/blob/master/parallel-consumer-core/src/test-integration/java/io/confluent/parallelconsumer/integrationTests/LoadTest.java)

Aka loading a topic with 4k messages and consuming with sleeps to represent database accesses
The performance results are:
- Normal Java Kafka Client accessed via Python: 0.43 seconds
- Parallel Consumer accessed via Python: 2.41 seconds
- Normal Python Kafka Client: 0.02 seconds

So in this use case parallel consumer is the worst option

astubbs: You can't do what PC does here, with the normal clients. PC enforces ordering by key, the other clients do not. What happens when you change line 70 from:

        .ordering(ParallelConsumerOptions.ProcessingOrder.KEY)


to:

        .ordering(ParallelConsumerOptions.ProcessingOrder.UNORDERED)

?

parallel-consumer-python/pyallel_consumer/pom.xml
astubbs: Ah, I thought we had some cp deps somewhere else, sharing a cp version number - turns out we don't. The only place we use cp version number is for the test container, inside the source. Ideally it'd be shared, but given this is outdated - all good.

astubbs (commented):
lgtm

parallel-consumer-python/tests/load/test_java_clients.py
Robbie-Palmer: That run, with `UNORDERED` I got

- Normal Java Kafka Client accessed via Python: 0.72 seconds
- Parallel Consumer accessed via Python: 2.70 seconds
- Normal Python Kafka Client: 0.16 seconds

So large variability in the others, but PC via Python in the same ballpark as before

astubbs: Changing a setting in PC, isn't going to affect the timing of the tests not using PC. Ping me on slack, and lets go through this together instead of like this.

astubbs: oh, I just noticed. Are the other two tests also using transactions? PC is also set to use transactions in the python test.
I see it's concurrency is only set to 3 as well - try 100, or 500.

astubbs: BTW, the test I was using in the weekend to test out the new engine speeds is `MultiInstanceHighVolumeTest`.

astubbs: Also, try it with the larger volumes of messages. Like several million.

astubbs: The async commit mode is probably what you want, unless the other tests are explicitly stopping to commit.

astubbs: And - try comparing it to using vanilla pc.

Robbie-Palmer: Yeh there's too many settings to tweak that I don't know enough about
Will ping you :)

Robbie-Palmer:
@astubbs this is now published to [PyPi](https://pypi.org/project/pyallel-consumer/) via `mvn deploy`
So CD is setup to publish on release

With it on PyPi it can now be installed normally with `pip install pyallel-consumer` and included in other projects dependencies etc.
I think this is good to go, as I don't think the one unresolved comment about the load tests is a blocker, more a nice to have?

eddyv:
Closing - Stale.

Issue #409, opened 2022-09-12
Parallel Consumer keeps committing the old offset after OFFSET_OUT_OF_RANGE in auto.offset.reset = latest


leonoah86:
Hi @astubbs, 
Thanks for this great library. Recently, I have been consistently seeing the consumer lag on some partitions of the topic when offset become out of range for that partition. This happens even with latest PC version - 0.5.2.2. I'm using KEY ordering.

This might be related to #352. Sample logs below.

- Last committed offset is 86321 and it's no longer available on broker due to retention policy.
- And it gets reset to latest offset - 88617

`12-09-2022 11:48:42.904 TraceId/SpanId: / [pc-broker-poll] INFO  org.apache.kafka.clients.consumer.internals.Fetcher.handleOffsetOutOfRange - [Consumer clientId=my-topic-consumer-client, groupId=my-topic-consumer] Fetch position FetchPosition{offset=86321, offsetEpoch=Optional.empty, currentLeader=LeaderAndEpoch{leader=Optional[localhost:6668 (id: 1 rack: null)], epoch=596}} is out of range for partition my-topic-4, resetting offset
12-09-2022 11:48:43.148 TraceId/SpanId: / [pc-broker-poll] INFO  org.apache.kafka.clients.consumer.internals.SubscriptionState.maybeSeekUnvalidated - [Consumer clientId=my-topic-consumer-client, groupId=my-topic-consumer] Resetting offset for partition my-topic-4 to position FetchPosition{offset=88617, offsetEpoch=Optional.empty, currentLeader=LeaderAndEpoch{leader=Optional[localhost:6668 (id: 1 rack: null)], epoch=596}}.
..`

New message at offset 88618 get consumed by the consumer

`12-09-2022 11:49:32.550 TraceId/SpanId: c65c59c9b8c75d88/b26a1c4eea95894d [pc-pool-3-thread-1] INFO  my.test.kafka.consumer.MyConsumer.consume - event=PROCESSING_COMPLETED, Event details: [id=cfd1d7e0-324d-11ed-8e3e-99871a3f782d, name=Event started, partition=4, offset=88618]
[...]

PR #445, opened 2022-10-19
refactor #409: Clarify truncation code

astubbs:
blocked by:
- #449

### DIFF
--- a/parallel-consumer-core/src/test-integration/java/io/confluent/parallelconsumer/integrationTests/state/PartitionStateCommittedOffsetIT.java
+++ b/parallel-consumer-core/src/test-integration/java/io/confluent/parallelconsumer/integrationTests/state/PartitionStateCommittedOffsetIT.java
@@ -4,7 +4,7 @@ package io.confluent.parallelconsumer.integrationTests.state;
  * Copyright (C) 2020-2022 Confluent, Inc.
  */
 
-import com.google.common.truth.Truth;
+import com.google.common.truth.StringSubject;
 import io.confluent.csid.utils.JavaUtils;
 import io.confluent.csid.utils.ThreadUtils;
 import io.confluent.parallelconsumer.ManagedTruth;
@@ -31,7 +31,6 @@ import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.EnumSource;
 import org.testcontainers.containers.KafkaContainer;
 import org.testcontainers.shaded.org.awaitility.Awaitility;
-import pl.tlinkowski.unij.api.UniLists;
 import pl.tlinkowski.unij.api.UniMaps;
 import pl.tlinkowski.unij.api.UniSets;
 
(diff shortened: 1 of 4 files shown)


parallel-consumer-core/src/test-integration/java/io/confluent/parallelconsumer/integrationTests/state/PartitionStateCommittedOffsetIT.java
```diff
@@ -69,6 +72,8 @@ class PartitionStateCommittedOffsetIT extends BrokerIntegrationTest<String, Stri
 
     private OffsetResetStrategy offsetResetStrategy = DEFAULT_OFFSET_RESET_POLICY;
 
+    private ParallelEoSStreamProcessor<String, String> activePc;
```
rkolesnev, line 75: consider returning PC instance from runPcCheckStartIs method instead of using handover through that private var.
It's only used in one test - to check failure cause - on offset reset policy NONE.

parallel-consumer-core/src/test-integration/java/io/confluent/parallelconsumer/integrationTests/state/PartitionStateCommittedOffsetIT.java
```diff
@@ -345,34 +379,21 @@ private List<PollContext<String, String>> runPcUntilOffset(OffsetResetStrategy o
...
+                log.debug("Succeeded up to: {}", succeededOffsets.last().offset());
             }
+            log.debug("Consumed up to {}", seenOffsets.last().offset());
+
+            var sorted = new ArrayList<>(seenOffsets);
+            Collections.sort(sorted, Comparator.comparingLong(PollContext::offset));
+            return sorted;
         } finally {
             try {
                 tempPc.close(); // close manually in this branch only, as in other branch it crashes
```
rkolesnev, line 399: not sure what that comment supposed to mean - leftover from branching logic ?

rkolesnev: any reason you are suppressing exception thrown by close()? 
is it possible for PC to fail close with exception when you are not expecting it or is it to prevent exception on close() overshadowing exception during execution (due to NONE offset reset policy / offset not present on poll)?  
Just thinking - if you only want to execute close when PC is not already closed or failed - cant you do same check as above on failFast i.e.
```
finally {
      if(!tempPc.isClosedOrFailed()) {
              temPc.close();
      }
}
```

rkolesnev (approved):
couple of comments on tests, otherwise LGTM

parallel-consumer-core/src/test-integration/java/io/confluent/parallelconsumer/integrationTests/state/PartitionStateCommittedOffsetIT.java
rkolesnev: wrong method referenced in comment - correct one is runPcUntilOffset - which already has a return type. So it really up to you to either wrap returned value (as Pair /Tuple / custom class) to hold both existing return value and in addition to it the ParallelEoSStreamProcessor or just leave as is...

PR #467, opened 2022-11-02
DP-9370 - use cc-service-bot to manage Semaphore project

cchristous:
This PR onboards the repo to cc-service-bot for managing its Semaphore project via code.

### DIFF
--- a/service.yml
+++ b/service.yml
@@ -9,4 +9,7 @@ github:
     enable: true
     repo_name: confluentinc/parallel-consumer
 semaphore:
-    enable: false
+  enable: true
+  pipeline_enable: false
+  triggers: []
+  branches: []


astubbs (changes requested):
Given that this project isn't built by semaphore and isn't allowed to be as it's public, afaik, why don't we just remove all the semaphore traces?

It's also currently disabled.

@confluentinc/tools

cchristous:
> Given that this project isn't built by semaphore and isn't allowed to be as it's public, afaik, why don't we just remove all the semaphore traces?
> 
> It's also currently disabled.
> 
> @confluentinc/tools

Sure, can you confirm for me that I can delete the https://confluentinc.semaphoreci.com/projects/parallel-consumer project? If so, I will delete it and close this PR.

Issue #71, opened 2021-01-20
Health-checks


JorgenRingen:
Kafka-streams exposes a method for retrieving the current state: `org.apache.kafka.streams.KafkaStreams#state`. 
We use this for implementing health-checks. If `state` is in `created, rebalancing or running` the application is considered healthy. Otherwise the app is considered unhealthy and the application would normally be stopped/restarted.

In PC I've implemented a health-check querying `io.confluent.parallelconsumer.ParallelEoSStreamProcessor#isClosedOrFailed == false`. 

Would a different, more "healthcheck-like" abstraction be smart? Perhaps on the `io.confluent.parallelconsumer.ParallelConsumer` interface? The current method requires a cast to ParallelEoSStreamProcessor.

(not critical!)

astubbs:
Do you have any thoughts on what metrics you might consider would indicate health? One major aspect is "progress" through the work queues. Described here: https://github.com/confluentinc/parallel-consumer/issues/34
Time since last progress made perhaps. Or time since last committable offset shift - per partition?

JorgenRingen:
In kafka-streams any leaked exception from user-code or any "permanent internal errors" causes the state to change to `Error -> Pending Shutdown -> Not Running`. We use this for our (aggregate) application health-checks and polls KS.state. If KS is in a unhealthy state the app is typically restarted by container-platform. Sometimes a restart can make the app recover and other times there's a poison-message being consumed that causes a restart-loop and needs manual intervention.

So maybe exposing something like `state` on `ParallelConsumer` interface for application-healthchecks might be a good start and any "shutdown"-scenarios in #34 would change the state to "error".

I'm doing ParallelEosStreamProcessor#isClosedOrFailed now and it works ok, but maybe a more uniform interface wold be smart down the road.

Current spring-boot-actuator-pc-healthcheck :) 
```
@Component
class ParallelConsumerHealthIndicator(
    parallelConsumer: ParallelStreamProcessor<String, Any>,
) : HealthIndicator {
    
    private val logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass())

    val parallelEosStreamProcessor = parallelConsumer as ParallelEoSStreamProcessor<String, Any> // notice cast

    override fun health(): Health {
        return if (!parallelEosStreamProcessor.isClosedOrFailed) {
            Health.up().build()
        } else {
            if (parallelEosStreamProcessor.failureCause != null) {
                logger.warn(
[...]

PR #485, opened 2022-11-29
Feature/health metrics micrometer [WIP]

nachomdo:
Fixes: 
* https://github.com/confluentinc/parallel-consumer/issues/71

Seeking some feedback associated to the following areas: 

 * **zero dependencies policy**: should we add MicroMeter SPI as a dependency to parallel-consumer-core or extract the PCMetricsTracker into a separate parallel-consumer-metrics project? another option would be to send it [upstream](https://github.com/micrometer-metrics/micrometer/tree/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder) with the rest of the existing binders for well-known libraries
  
 * **Devex/usage**: this approach leaves in the hands of the developer the register setup and binding. Another approach could be to pass the register as a `ParallelConsumerOption` and take care of the binder inside of `AbstractParallelEoSStreamProcessor`. 
 
 * **Healthchecks**: It seems not possible to propagate a [string as a Gauge value](https://github.com/micrometer-metrics/micrometer/issues/1033), so I'm using a tag to decorate the numerical value associated with PCState (e.g. running, paused, closed, etc)

This current draft only binds a subset of the metrics. Happy to keep iterating on it if the approach is okay with you 😊  

### Checklist

- [ ] Changelog

Replaces / Makes redundant:
- #464


astubbs:
Is it possible to add an integration tests that spins up Prometheus and Grafana in a test container, to test out the metric integration?

astubbs:
From GPTChat

How do you configure Prometheus describe micro meter?

To configure Prometheus to scrape metrics from a micro meter, you will need to add the appropriate configuration to the Prometheus server. This typically involves adding a new scrape configuration that specifies the target for the scrape, the metrics to collect, and any additional settings or options.

For example, to scrape metrics from a micro meter, you might add a configuration like the following to the Prometheus server:

Copy code
scrape_configs:
  - job_name: "micro-meter"
    scrape_interval: 15s
    metrics_path: /metrics
    static_configs:
      - targets: ["localhost:8080"]
This configuration specifies a new scrape job named "micro-meter", with a scrape interval of 15 seconds. It specifies that Prometheus should scrape metrics from the "/metrics" endpoint on the target "localhost:8080".

Once this configuration is added to the Prometheus server, it will begin scraping metrics from the micro meter at the specified interval. The scraped metrics will be stored in the Prometheus database, and they can be queried and visualized using tools like Grafana.

Overall, configuring Prometheus to scrape metrics from a micro meter involves adding the appropriate configuration to the Prometheus server, specifying the target, metrics path, and other settings for the scrape. This can provide valuable insights into the performance and health of your application, and it can help you to monitor and alert on potential issues and problems.

astubbs:
How do you set up grafana to visualise Prometheus data?

To set up Grafana to visualize Prometheus data, you will need to add a Prometheus data source to Grafana and configure it to connect to your Prometheus server. This will allow Grafana to query the Prometheus server and retrieve the metrics that have been collected by Prometheus.

To add a Prometheus data source to Grafana, follow these steps:

Log in to Grafana and go to the "Data Sources" page.
Click the "Add data source" button.
Select "Prometheus" from the list of data source types.
Enter a name for the data source, such as "Prometheus Server".
In the "URL" field, enter the URL of your Prometheus server, including the port number (e.g., "http://localhost:9090").
Leave the "Access" field set to "Direct".
Click the "Save & Test" button to save the data source and test the connection to the Prometheus server.
Once the Prometheus data source has been added to Grafana, you can use it to create dashboards and visualizations that display the metrics collected by Prometheus. To do this, follow these steps:

Go to the "Dashboards" page in Grafana and click the "New dashboard" button.
Select "Graph" from the list of panel types.
In the "Metrics" tab, select the Prometheus data source that you added earlier.
Enter a Prometheus query in the "Query" field to retrieve the metrics that you want to visualize.
Use the other settings and options in the "Metrics" tab to customize the appearance and behavior of the graph.
Click the "Save" button to save the dashboard and the graph.
Once the dashboard and graph have been saved, Grafana will automatically refresh the graph with the latest data from Prometheus, and it will display the metrics that you queried. You can then use the dashboard and graph to monitor and analyze the metrics collected by Prometheus, and you can use the other features and capabilities of Grafana to customize and enhance the visualizations.

Overall, setting up Grafana to visualize Prometheus data involves adding a Prometheus data source to Grafana and configuring it to connect to your Prometheus server. This will allow you to use Grafana to create dashboards and visualizations that display the metrics collected by Prometheus, and it will provide valuable insights into the performance and health of your application.



Antony Stubbs
Are there TestContainers for grafana and Prometheus? And are there ones that come already setup?

Yes, there are TestContainers for Grafana and Prometheus that you can use to spin up containers for these tools in your integration tests. TestContainers is a Java library that provides a simple and lightweight way to run containers in tests, and it allows you to easily spin up containers for tools like Grafana and Prometheus in your integration tests.

TestContainers provides a pre-configured container for Grafana, as well as a container for Prometheus. These containers come with the appropriate settings and configuration already applied, so you can spin them up in your tests without having to configure them manually.

To use TestContainers to spin up Grafana and Prometheus containers in your tests, you will need to add the TestContainers library as a dependency to your project, and you will need to import the appropriate containers from the TestContainers library. For example, you can use the following code to import the Grafana and Prometheus containers from TestContainers:

Copy code
import org.testcontainers.containers.GraphiteContainer;
import org.testcontainers.containers.PrometheusContainer;
Once the Grafana and Prometheus containers are imported, you can use them to spin up the containers in your tests. For example, you might use the following code to spin up the Grafana and Prometheus containers in your tests:

Copy code
// Create the Grafana and Prometheus containers
GraphiteContainer grafana = new GraphiteContainer();
PrometheusContainer prometheus = new PrometheusContainer();

// Start the containers
grafana.start();
prometheus.start();

// Use the containers in your tests...

// Stop the containers when you are done
grafana.stop();
prometheus.stop();
This code will spin up the Grafana and Prometheus containers in your tests, and it will start the containers with the appropriate settings and configuration. You can then use the containers in your tests, and you can stop the containers when you are done.

Overall, TestContainers provides pre-configured containers for Grafana and Prometheus that you can use to spin up these tools in your integration tests. This can provide a convenient and lightweight way to test the integration between your application and these tools, and it can help you to ensure that the metrics exposed by your application are being collected and stored correctly by Prometheus.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ShardManager.java
```diff
@@ -37,6 +38,7 @@
  *
  * @author Antony Stubbs
  */
+// metrics: number of queues, average queue length
 @Slf4j
 public class ShardManager<K, V> {
```
astubbs, line 43: slow work gauge of some sort - how many records are "slow", average wait time, something like that. average size of the slow work collection every time we try to take work.

cc @Ehud-Lev-Forter

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/PCMetrics.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.Optional;
+
+/**
+ * Metrics model for Parallel Consumer
+ *
+ * @author Antony Stubbs
+ */
+@Value
+@SuperBuilder(toBuilder = true)
+public class PCMetrics {
```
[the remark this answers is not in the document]
nachomdo: Debug-level metrics like we have in Kafka Streams could be a good approach. Also, MicroMeter allows filtering when setting up the registry so developers can selectively drop metrics. 

```
registry.config()
    .meterFilter(MeterFilter.ignoreTags("too.much.information"))
    .meterFilter(MeterFilter.denyNameStartsWith("jvm"));
```

nachomdo:
👋  @astubbs , I've just added the Prometheus/Grafana plumbing and IntegrationTests based on TestContainers that we agreed on and wired the MeterRegistry dependency using PCModule as per your example in your remote. 

On a different note, I also played with some potential Grafana visualizations and [documented](https://github.com/confluentinc/parallel-consumer/pull/485/commits/cc5107c33e0ca000778928d6d85a68790ec2e057#diff-118ae53707b50d141678d0a6949a1fee60932b8eb3d391bfc240edff7b35fca0) most of the metrics I have been looking at. 

image
image
image

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/PCMetrics.java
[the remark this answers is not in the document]
cykl: > Also, MicroMeter allows filtering when setting up the registry so developers can selectively drop metrics.

MeterFilter are powerful but annoying. IMHO It would be much more pleasant to be able to enable "basics" and "detailed" metrics using specifics binder or by configuring the binder. One can still use MeterFilters in specific circumstance but they shouldn't be required by the happy path. 

I think per shard metrics are not possible with key level parallelism and a non trivial amount of keys.

1.  IIRC the memory footprint of a timer is in the 2..8kB range depending if percentile / SLO / histogram are enabled. It's likely bearable for most use case but not unremarkable.
2. I think that most TSDB won't be able to handle such cardinality, at least in a cost effective way. 

I wouldn't enable them by default but look for a way to be able to spot anomalies using low cardinality meters, logs or any other means.


Per shard metrics become impossible 

 
It would be nice to not have to rely on metrics filtering to opt-out from shard level metrics, I find it annoying and inefficient compared to be just bind desired meters. Splitting PCMetricsTracker or making it configurable may be one way to do it.

cykl:
> and  most of the metrics I have been looking at.

I see discrepancies between this document and PcMetricsTracker. Which one should be considered as the reference?

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/PCMetrics.java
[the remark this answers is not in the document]
astubbs: Thank you so much for input @cykl !

> I think per shard metrics are not possible with key level parallelism and a non trivial amount of keys.

Do you mean all types of shard metrics? or only timers? or ..?

astubbs:
> 👋 @astubbs , I've just added the Prometheus/Grafana plumbing and IntegrationTests based on TestContainers that we agreed on and wired the MeterRegistry dependency using PCModule as per your example in your remote.

OMG these looks amazing! Can't wait to try it out!

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -272,6 +330,15 @@ protected AbstractParallelEoSStreamProcessor(ParallelConsumerOptions<K, V> newOp
...
             this.committer = this.brokerPollSubsystem;
         }
+
+        new ExecutorServiceMetrics(workerThreadPool, "pc-worker-executor", Collections.emptyList())
+                .bindTo(module.meterRegistry());
+
+        this.userProcessingTimer = io.micrometer.core.instrument.Timer.builder("pc.user.function.processing.time")
+                .description("user function processing time")
+                .publishPercentileHistogram(true)
+                .publishPercentiles(0.5,0.95,0.99,0.999)
```
astubbs, line 340: extract constant? will be quite typical right?

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -1132,79 +1199,61 @@ private void updateLastCommitCheckTime() {
...
+        getWm().addMetricsTo(metrics);
 
-            resultsFromUserFunction = usersFunction.apply(context);
+        metrics.functionTimer(userProcessingTimer);
+        metrics.successCounter(successCounter);
 
-            for (final WorkContainer<K, V> kvWorkContainer : workContainerBatch) {
-                onUserFunctionSuccess(kvWorkContainer, resultsFromUserFunction);
-            }
+        // should not create objects in metrics calculator method
```
astubbs, line 1218: resolve

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -1132,79 +1199,61 @@ private void updateLastCommitCheckTime() {
...
-     * Run the supplied function.
+     * Gather metrics from the {@link ParallelConsumer} and for monitoring.
      */
-    protected <R> List<ParallelConsumer.Tuple<ConsumerRecord<K, V>, R>> runUserFunction(Function<PollContextInternal<K, V>, List<R>> usersFunction,
-                                                                                        Consumer<R> callback,
-                                                                                        List<WorkContainer<K, V>> workContainerBatch) {
-        // call the user's function
-        List<R> resultsFromUserFunction;
-        PollContextInternal<K, V> context = new PollContextInternal<>(workContainerBatch);
+    public PCMetrics calculateMetrics() {
```
astubbs, line 1208: does it make sense to have this manual call still? what should our "health" end point look like?

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -1132,79 +1199,61 @@ private void updateLastCommitCheckTime() {
...
-            // fail or succeed, either way we're done
-            for (var kvWorkContainer : workContainerBatch) {
-                addToMailBoxOnUserFunctionSuccess(context, kvWorkContainer, resultsFromUserFunction);
-            }
-            log.trace("User function future registered");
+            // need to add closables to close system on shutdown
+            {
+                new JvmGcMetrics().bindTo(metricsRegistry);
+                new KafkaClientMetrics(consumer).bindTo(metricsRegistry);
+                // can bind to wrapper?
```
astubbs, line 1229: resolve

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -1132,79 +1199,61 @@ private void updateLastCommitCheckTime() {
...
-            var cause = e.getCause();
-            String msg = msg("Exception caught in user function running stage, registering WC as failed, returning to" +
-                    " mailbox. Context: {}", context, e);
-            if (cause instanceof PCRetriableException) {
-                log.debug("Explicit " + PCRetriableException.class.getSimpleName() + " caught, logging at DEBUG only. " + msg, e);
-            } else {
-                log.error(msg, e);
-            }
+                // we don't use an admin client?
+                // new KafkaClientMetrics(admin).bindTo(metricsRegistry);
```
astubbs, line 1233: resolve

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -1132,79 +1199,61 @@ private void updateLastCommitCheckTime() {
...
 
-            for (var wc : workContainerBatch) {
-                wc.onUserFunctionFailure(e);
-                addToMailbox(context, wc); // always add on error
+                // can't for logback?
+                //new LogbackMetrics().bindTo(metricsRegistry);
```
astubbs, line 1236: resolve

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -1132,79 +1199,61 @@ private void updateLastCommitCheckTime() {
...
         }
+
+        return metrics.build();
     }
 
     protected void addToMailBoxOnUserFunctionSuccess(PollContextInternal<K, V> context, WorkContainer<K, V> wc, List<?> resultsFromUserFunction) {
         addToMailbox(context, wc);
     }
 
+    Counter successCounter = Counter
```
astubbs, line 1247: private?

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -1303,4 +1352,111 @@ public void resumeIfPaused() {
...
+        private boolean isWorkResult() {
+            return workContainer != null;
+        }
+
+        private boolean isNewConsumerRecords() {
+            return !isWorkResult();
+        }
+
+
+//        CompositeMeterRegistry compositeRegistry = new CompositeMeterRegistry();
```
astubbs, line 1377: cleanup

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -1303,4 +1352,111 @@ public void resumeIfPaused() {
...
+
+        public static <K, V> ControllerEventMessage<K, V> of(WorkContainer<K, V> work) {
+            return new ControllerEventMessage<K, V>(work, null);
+        }
+    }
+
+    /**
+     * todo docs
+     */
+    public PCMetrics calculateMetricsWithIncompletes() {
```
astubbs, line 1447: as above, what should our health end point look like, and does it makes sense to have direct access like this?

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/BrokerPollSystem.java
```diff
@@ -33,6 +34,7 @@
  * @param <K>
  * @param <V>
  */
+// metrics: avg time spent polling, number of paused partitions?
```
astubbs, line 37: resolve

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -44,11 +55,16 @@
 import static lombok.AccessLevel.PROTECTED;
 
 /**
+ * @author Antony Stubbs
  * @see ParallelConsumer
  */
+// metrics: avg time spend committing, avg user function running time, avg controller loop time
```
astubbs, line 61: resolve

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/PCModule.java
```diff
@@ -22,6 +23,16 @@
  * @author Antony Stubbs
  */
 public class PCModule<K, V> {
+    public static final String PARALLEL_CONSUMER_PREFIX = "parallel-consumer.";
```
astubbs, line 26:
```suggestion

    public static final String PARALLEL_CONSUMER_PREFIX = "parallel-consumer.";
```

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/PCModule.java
```diff
@@ -22,6 +23,16 @@
...
+
+    MeterRegistry meterRegistry;
```
astubbs, line 28:
```suggestion
    protected MeterRegistry meterRegistry;
```

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/offsets/OffsetEncoder.java
```diff
@@ -17,6 +17,7 @@
  *
  * @author Antony Stubbs
  */
+// metrics: avg offsets mapped per bit, average encoded size, avg time to encode,
```
astubbs, line 20: resolve

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/offsets/OffsetMapCodecManager.java
```diff
@@ -100,6 +108,12 @@ public static HighestOffsetAndIncompletes of() {
     // todo remove consumer #233
     public OffsetMapCodecManager(PCModule<K, V> module) {
         this.module = module;
+
+        avgEncodingTimeMeter = Timer.builder("pc.avg.encoding.time")
+                .description("Average time spend encoding offsets")
+                .publishPercentileHistogram(true)
+                .publishPercentiles(0.5, 0.95, 0.99, 0.999)
```
astubbs, line 115: DRY

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/offsets/OffsetMapCodecManager.java
```diff
@@ -100,6 +108,12 @@ public static HighestOffsetAndIncompletes of() {
     // todo remove consumer #233
     public OffsetMapCodecManager(PCModule<K, V> module) {
         this.module = module;
+
+        avgEncodingTimeMeter = Timer.builder("pc.avg.encoding.time")
+                .description("Average time spend encoding offsets")
```
astubbs, line 113: convention is start with upper case, or?

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/offsets/OffsetMapCodecManager.java
```diff
@@ -185,6 +199,14 @@ String serialiseIncompleteOffsetMapToBase64(long baseOffsetForPartition, Partiti
         String b64 = OffsetSimpleSerialisation.base64(compressedEncoding);
         return b64;
     }
+    private Timer avgEncodingTimeMeter;
```
astubbs, line 202: lets move all the fields to the top

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/offsets/OffsetMapCodecManager.java
```diff
@@ -207,22 +229,26 @@ byte[] encodeOffsetsCompressed(long baseOffsetForPartition, PartitionState<K, V>
 
         OffsetSimultaneousEncoder simultaneousEncoder = null;
         try {
-            simultaneousEncoder = new OffsetSimultaneousEncoder(baseOffsetForPartition, highestSucceeded, incompleteOffsets);
-            simultaneousEncoder.invoke();
-        } catch (Exception e) {
+            simultaneousEncoder =  new OffsetSimultaneousEncoder(baseOffsetForPartition, highestSucceeded, incompleteOffsets);
+            avgEncodingTimeMeter.recordCallable(simultaneousEncoder::invoke);
```
astubbs, line 233: slick

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/offsets/OffsetMapCodecManager.java
```diff
@@ -207,22 +229,26 @@ byte[] encodeOffsetsCompressed(long baseOffsetForPartition, PartitionState<K, V>
...
+         } catch (Exception e) {
             throw new InternalRuntimeException("Error encoding offsets", e);
         }
 
         //
         if (forcedCodec.isPresent()) {
             var forcedOffsetEncoding = forcedCodec.get();
             log.debug("Forcing use of {}, for testing", forcedOffsetEncoding);
+            getCounterMeterFor("pc.encoding.usage", forcedOffsetEncoding).increment();
```
astubbs, line 242: constant?

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/offsets/OffsetMapCodecManager.java
```diff
@@ -207,22 +229,26 @@ byte[] encodeOffsetsCompressed(long baseOffsetForPartition, PartitionState<K, V>
...
+
             Map<OffsetEncoding, byte[]> encodingMap = simultaneousEncoder.getEncodingMap();
             byte[] bytes = encodingMap.get(forcedOffsetEncoding);
             if (bytes == null)
                 throw new NoEncodingPossibleException(msg("Can't force an encoding that hasn't been run: {}", forcedOffsetEncoding));
             return simultaneousEncoder.packEncoding(new EncodedOffsetPair(forcedOffsetEncoding, ByteBuffer.wrap(bytes)));
         } else {
+            getCounterMeterFor("pc.encoding.usage", simultaneousEncoder.sortedEncodings.first().getEncoding())
```
astubbs, line 250: DRY

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/offsets/RunLengthEncoder.java
```diff
@@ -6,6 +6,9 @@
 
 import io.confluent.csid.utils.MathUtils;
 import io.confluent.csid.utils.Range;
+import io.micrometer.core.instrument.DistributionSummary;
+import io.micrometer.core.instrument.Metrics;
+import io.micrometer.core.instrument.binder.BaseUnits;
```
astubbs, line 11: unused imports?

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/PartitionState.java
```diff
@@ -35,6 +39,10 @@
  * @see PartitionStateManager
  */
 // todo class becoming large - possible to extract some functionality?
+// metrics: current diff between highest succeeded and committable offset (offset tracked range / state size), blocked
```
astubbs, line 42: resolve

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/PartitionState.java
```diff
@@ -156,15 +164,30 @@
...
+
         this.tp = topicPartition;
         this.partitionsAssignmentEpoch = newEpoch;
 
         initStateFromOffsetData(offsetData);
+
+        ratioOfPayloadUsed = DistributionSummary.builder("pc.payload.ratio.used")
+                .description("ratio between offset metadata payload size and offsets encoded")
+                .publishPercentileHistogram(true)
+                .publishPercentiles(0.5, 0.95, 0.99, 0.999)
```
astubbs, line 184: DRY

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/PartitionState.java
```diff
@@ -156,15 +164,30 @@
...
+                .register(module.meterRegistry());
+        ratioOfMetadataSpaceUsed = DistributionSummary.builder("pc.metadata.space.used")
```
astubbs, line 186: patterns looks very similar, create a factor that sets all the defaults?

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/PartitionState.java
```diff
@@ -444,9 +469,12 @@ private Optional<String> tryToEncodeOffsets() {
 
         try {
             // todo refactor use of null shouldn't be needed. Is OffsetMapCodecManager stateful? remove null #233
-            OffsetMapCodecManager<K, V> om = new OffsetMapCodecManager<>(null);
+            OffsetMapCodecManager<K, V> om = new OffsetMapCodecManager<>(this.module);
             long offsetOfNextExpectedMessage = getOffsetToCommit();
+            var offsetRange = getOffsetHighestSucceeded() - offsetOfNextExpectedMessage;
             String offsetMapPayload = om.makeOffsetMetadataPayload(offsetOfNextExpectedMessage, this);
+            ratioOfPayloadUsed.record(offsetMapPayload.length() / (double) offsetRange);
```
astubbs, line 476: audit

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/PartitionStateManager.java
```diff
@@ -32,6 +32,7 @@
  * @author Antony Stubbs
  * @see PartitionState
  */
+// metrics: assigned partitions and their epochs, number of assigned partitions,
```
astubbs, line 35: resolve

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ProcessingShard.java
```diff
@@ -26,7 +27,9 @@
  * Models the queue of work to be processed, based on the {@link ProcessingOrder} modes.
  *
  * @author Antony Stubbs
+ * @see ShardManager
  */
+// metrics: queue length, queue time, avg time spend in queue
```
astubbs, line 32: resolve

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/RemovedPartitionState.java
```diff
@@ -42,7 +44,7 @@
     public static final int NO_EPOCH = -1;
 
     public RemovedPartitionState() {
-        super(NO_EPOCH, null, null, OffsetMapCodecManager.HighestOffsetAndIncompletes.of());
+        super(NO_EPOCH, new PCModule<>(ParallelConsumerOptions.<K, V>builder().build()), null, OffsetMapCodecManager.HighestOffsetAndIncompletes.of());
```
astubbs, line 47: shouldn't be building new modules - make constructor private? restrict somehow? have had problems with multiple modules in the past sneaking their way in

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ShardManager.java
```diff
@@ -37,11 +38,13 @@
  *
  * @author Antony Stubbs
  */
+// metrics: number of queues, average queue length
```
astubbs, line 41: resolve

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/WorkManager.java
```diff
@@ -58,19 +60,28 @@
...
+
     public WorkManager(PCModule<K, V> module,
                        DynamicLoadFactor dynamicExtraLoadFactor) {
+        this.module = module;
         this.options = module.options();
         this.dynamicLoadFactor = dynamicExtraLoadFactor;
         this.sm = new ShardManager<>(module, this);
         this.pm = new PartitionStateManager<>(module, sm);
+
+        inflightRecordsMeter = Gauge.builder("pc.records.inflight", this::getNumberRecordsOutForProcessing)
```
astubbs, line 81: dry prefix

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/WorkManager.java
```diff
@@ -58,19 +60,28 @@
...
+                .register(module.meterRegistry());
+        waitingRecordsMeter = Gauge.builder("pc.records.waiting", this::getNumberOfWorkQueuedInShardsAwaitingSelection)
```
astubbs, line 83: should depend on getNumberOfWorkQueuedInShardsAwaitingSelection performance fixes from GPT branch?

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/PCMetrics.java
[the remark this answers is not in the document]
cykl: Likely both. 

Timers are worst offenders because each timer generates multiple time-series in the TSDB (ex. count, totalTime, maxTime + buckets for histograms etc.) while a gauge or a counter creates only one.

But if you have many shards, even a single gauge with a shard label could be problematic.

It is often recommended to not use user supplied input for tags values and to be mindful of the cardinality.

Micrometer:
> Beware of the potential for tag values coming from user-supplied sources to blow up the cardinality of a metric. You should always carefully normalize and add bounds to user-supplied input. Sometimes, the cause is sneaky. Consider the URI tag for recording HTTP requests on service endpoints. If we do not constrain 404’s to a value like NOT_FOUND, the dimensionality of the metric would grow with each resource that cannot be found.

Prometheus:
> Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.

My understanding is that with key parallelism, shardId = key which can be a very large set of values. In this case you will likely disable shard level metrics because they are too expensive. 

vaguely related question: is PC able to deal with a large number of keys (ex. millions) with key level parallelism? My understanding is that you have a per shard memory overhead and offset maps are going to contains many entries possibly exceeding the 4KB limit. Is there any benchmark / feedback available about that?

parallel-consumer-examples/parallel-consumer-example-core/pom.xml
```diff
@@ -35,8 +35,74 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.micrometer</groupId>
+            <artifactId>micrometer-registry-prometheus</artifactId>
+            <version>1.8.2</version>
+        </dependency>
+
+        <dependency>
```
astubbs, line 44: should not be needed

parallel-consumer-examples/parallel-consumer-example-core/pom.xml
```diff
@@ -35,8 +35,74 @@
...
+            <groupId>org.testcontainers</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
+    <build>
+        <finalName>pc-core-metrics</finalName>
```
astubbs, line 75: doesn't seem right

parallel-consumer-examples/parallel-consumer-example-core/src/test/java/io/confluent/parallelconsumer/examples/core/CoreAppMetricsIntegrationTest.java
```diff
@@ -0,0 +1,100 @@
...
+import java.time.Duration;
+import java.util.Map;
+import java.util.Set;
+
+import static org.mockito.Mockito.when;
+import static pl.tlinkowski.unij.api.UniLists.of;
+
+@Slf4j
+@Testcontainers
+public class CoreAppMetricsIntegrationTest{
```
astubbs, line 38: does this work without needing the docker compose etc?

astubbs (changes requested):
big ticket discussion points sent on slack. Great work! getting there, now for the polish...

nachomdo:
:wave: @rkolesnev @astubbs  as per our discussion I have done the following changes: 

* Instrument classes surfacing structured events via an EventBus instead of using meters directly. If you prefer we can use the actor-based subsystem from PC. Just wanted to validate that it is aligned with @rkolesnev suggestion 😊  An example of raising the event is below 
https://github.com/confluentinc/parallel-consumer/blob/000bd95a4ebcf2bbe6bd39670ea344081aaa5be1/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/WorkManager.java#L158-L161 
and the processing is done in PCMetricsTracker 
https://github.com/confluentinc/parallel-consumer/blob/000bd95a4ebcf2bbe6bd39670ea344081aaa5be1/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/PCMetricsTracker.java#L92-L104

* Remove any docker-compose harness and Grafana dashboards to reduce PR noise

* Move metrics example using Prometheus into its own module 

* Merge master to apply these changes over the most up-to-date version 

If you are happy with the Structured Event approach I can continue by adding the remaining initial metrics documented [here](https://github.com/confluentinc/parallel-consumer/blob/000bd95a4ebcf2bbe6bd39670ea344081aaa5be1/METRICS.adoc).

parallel-consumer-examples/parallel-consumer-example-core/src/test/java/io/confluent/parallelconsumer/examples/core/CoreAppMetricsIntegrationTest.java
nachomdo: no docker-compose needed, just TestContainers

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
nachomdo: done

acktsap:
Any updates? I need this feature..

astubbs:
> Any updates? I need this feature..
> 
> FYI, i also made simple changes using micrometer [here](https://github.com/acktsap/parallel-consumer/pull/1) (I didn't know this PR exists).

Hi! I'm back online now, so this will really start moving. It's my top priority.

astubbs:
> Any updates? I need this feature..
> 
> FYI, i also made simple changes using micrometer [here](https://github.com/acktsap/parallel-consumer/pull/1) (I didn't know this PR exists).

BTW - we'd love to get your input into this PR?

PR #506, opened 2022-12-08
minor: docs: Fix chart links

astubbs:
This is an issue with GitHub starting to block embedding canvas images in rendered docs. Have contacted support to see what they say...
Seems it's due to Confluent locking down google doc access to org only.


eddyv:
Closing - Stale.

PR #518, opened 2022-12-20
Add missing @UtilityClass to utils class

acktsap:
Add missing `@UtilityClass` to utils

### Checklist

- [ ] Changelog

### DIFF
--- a/parallel-consumer-core/src/main/java/io/confluent/csid/utils/LoopingResumingIterator.java
+++ b/parallel-consumer-core/src/main/java/io/confluent/csid/utils/LoopingResumingIterator.java
@@ -5,7 +5,6 @@ package io.confluent.csid.utils;
  */
 
 import lombok.Getter;
-import lombok.experimental.UtilityClass;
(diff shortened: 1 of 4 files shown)


parallel-consumer-core/src/main/java/io/confluent/csid/utils/LoopingResumingIterator.java
```diff
@@ -5,6 +5,7 @@
  */
 
 import lombok.Getter;
+import lombok.experimental.UtilityClass;
```
astubbs, line 8: unused?
```suggestion
```

astubbs (changes requested):
Thanks!

parallel-consumer-core/src/main/java/io/confluent/csid/utils/LoopingResumingIterator.java
acktsap: Oh yes unused. I've removed import :)

PR #531, opened 2022-12-22
workerThreadPool is initialized lazily

pweso:
Hi,
We want to use Apache Kafka in our project, and the **parallel-consumer** library meets our expectations.
However, we would like to use it in a way that fits our architecture.

**What do I want to achieve?**
I would like to have control over the worker thread pool definition. We have our abstraction over thread pools that allows us to monitor them easily, and we have one place where we have all thread pool definitions.

**What steps have I taken?**
I created a class that inherits from `ParallelEoSStreamProcessor`, and I see that the `setupWorkerPool(int poolSize)` method in `AbstractParallelEoSStreamProcessor` is _protected_, so I overridden it and used my `ThreadPoolExecutor` implementation.
Nonetheless, I get a NPE because the `setupWorkerPool` method is called in the constructor of the `AbstractParallelEoSStreamProcessor` class, so it happens before the fields are assigned in my custom superclass (via the constructor). The solution is quite simple; that is, `workerThreadPool` is initialized lazily. The overridden abstract method will be called only the first time it is used, and then we can be sure that class, which inherits from `ParallelEoSStreamProcessor`, will be fully initialized.

What do you think about this change?

By the way, I'm just reading the javadoc of `ParallelConsumerOptions` regarding `maxConcurrency` and I see this sentence:

> When using getBatchSize(), this is over and above the batch size setting. So for example, a getMaxConcurrency() of 2 and a batch size of 3 would result in at most 15 records being processed at once.

I don't understand where 15 came from, could you clarify that?

### DIFF
--- a/parallel-consumer-core/pom.xml
+++ b/parallel-consumer-core/pom.xml
@@ -8,7 +8,7 @@
     <parent>
         <groupId>io.confluent.parallelconsumer</groupId>
         <artifactId>parallel-consumer-parent</artifactId>
-        <version>0.5.2.5-SNAPSHOT</version>
+        <version>0.5.2.6-SNAPSHOT</version>
     </parent>
 
     <modelVersion>4.0.0</modelVersion>
@@ -26,13 +26,13 @@
         <dependency>
             <groupId>com.github.luben</groupId>
             <artifactId>zstd-jni</artifactId>
-            <version>1.5.2-4</version>
+            <version>1.5.5-4</version>
             <scope>compile</scope>
         </dependency>
         <dependency>
             <groupId>org.xerial.snappy</groupId>
             <artifactId>snappy-java</artifactId>
-            <version>1.1.8.4</version>
+            <version>1.1.10.1</version>
             <scope>compile</scope>
         </dependency>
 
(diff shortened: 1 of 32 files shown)


parallel-consumer-core/pom.xml
```diff
@@ -35,6 +35,12 @@
             <version>1.1.8.4</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
```
astubbs, line 38: Quick comment - can you make this without adding dependencies?

astubbs (changes requested):

astubbs:
> > When using getBatchSize(), this is over and above the batch size setting. So for example, a getMaxConcurrency() of 2 and a batch size of 3 would result in at most 15 records being processed at once.

> I don't understand where 15 came from, could you clarify that?

Yeah that sounds complete wrong. From a quick look, it should read 6, but I'll take a look.

astubbs:
Your proposal seems good - just remove the Guava dep and we should be good to go.

pweso:
I didn't notice that there is zero~ dependencies in the features list. I removed Guava dependency and I wrote my own cached supplier implementation based on the information here: https://stackoverflow.com/a/35332514

parallel-consumer-core/pom.xml
pweso: Fixed

eddyv (approved):
lgtm.

rkolesnev (approved):

Issue #326, opened 2022-07-04
Error in onPartitionsAssigned in parallel consumer


milansanjeev:
Hi Team,

I want to use parallel consumer in one of over spring service to process kafka stream.
I am using core parallel-consumer-core 0.5.1.0 but getting below exception. We are secured kafka clsuter.

```

2022-07-04 15:14:49.224 app=opxhub-order-domain-bookings-stream environment=springEnvironment_IS_UNDEFINED ERROR 39138 --- [ pc-broker-poll] i.c.p.state.PartitionStateManager        : Error in onPartitionsAssigned

java.lang.RuntimeException: Unexpected magic: 1
	at io.confluent.parallelconsumer.offsets.OffsetEncoding.decode(OffsetEncoding.java:52)
	at io.confluent.parallelconsumer.offsets.EncodedOffsetPair.unwrap(EncodedOffsetPair.java:75)
	at io.confluent.parallelconsumer.offsets.OffsetMapCodecManager.decodeCompressedOffsets(OffsetMapCodecManager.java:229)
	at io.confluent.parallelconsumer.offsets.OffsetMapCodecManager.deserialiseIncompleteOffsetMapFromBase64(OffsetMapCodecManager.java:161)
	at io.confluent.parallelconsumer.offsets.OffsetMapCodecManager.deserialiseIncompleteOffsetMapFromBase64(OffsetMapCodecManager.java:151)
	at io.confluent.parallelconsumer.offsets.OffsetMapCodecManager.decodePartitionState(OffsetMapCodecManager.java:165)
	at io.confluent.parallelconsumer.offsets.OffsetMapCodecManager.lambda$loadPartitionStateForAssignment$0(OffsetMapCodecManager.java:128)
	at java.base/java.util.HashMap.forEach(HashMap.java:1425)
	at io.confluent.parallelconsumer.offsets.OffsetMapCodecManager.loadPartitionStateForAssignment(OffsetMapCodecManager.java:125)
	at io.confluent.parallelconsumer.state.PartitionStateManager.onPartitionsAssigned(PartitionStateManager.java:107)
	at io.confluent.parallelconsumer.state.WorkManager.onPartitionsAssigned(WorkManager.java:98)
	at io.confluent.parallelconsumer.internal.AbstractParallelEoSStreamProcessor.onPartitionsAssigned(AbstractParallelEoSStreamProcessor.java:360)
[...]

PR #537, opened 2023-01-13
improves: Error in onPartitionsAssigned in parallel consumer

nachomdo:
Improves: https://github.com/confluentinc/parallel-consumer/issues/326 

Adds an option to pass an invalid offset metadata error policy in case of accidentally or deliberately reusing a Kafka Streams consumer group. The default policy is to fail and stop the execution. 

### Checklist

- [ ] Changelog

### DIFF
--- a/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/offsets/OffsetMapCodecManager.java
+++ b/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/offsets/OffsetMapCodecManager.java
@@ -103,7 +103,9 @@ public class OffsetMapCodecManager<K, V> {
     // todo remove consumer #233
     public OffsetMapCodecManager(PCModule<K, V> module) {
         this.module = module;
-        this.errorPolicy = module.options().getInvalidOffsetMetadataPolicy();
+        if (module != null){
+            this.errorPolicy = module.options().getInvalidOffsetMetadataPolicy();
+        }
     }
 
     /**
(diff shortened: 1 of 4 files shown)


parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/offsets/OffsetMapCodecManager.java
```diff
@@ -100,6 +103,7 @@ public static HighestOffsetAndIncompletes of() {
     // todo remove consumer #233
     public OffsetMapCodecManager(PCModule<K, V> module) {
         this.module = module;
+        this.errorPolicy = module.options().getInvalidOffsetMetadataPolicy();
```
nachomdo, line 106: Not my favourite option this mixture of static and instance scope 😢

parallel-consumer-core/src/test/java/io/confluent/parallelconsumer/offsets/WorkManagerOffsetMapCodecManagerTest.java
eddyv, line 1: Should probably add a test for `deserialiseKafkaStreamsV2WithIgnoreErrorPolicy `

eddyv (changes requested):
Add one more test and LGTM.

parallel-consumer-core/src/test/java/io/confluent/parallelconsumer/offsets/WorkManagerOffsetMapCodecManagerTest.java
nachomdo: Thanks for the review @eddyv 🙇‍♂️  I've added the suggested test

eddyv (approved):
Lgtm

Issue #560, opened 2023-02-24
Feature suggestion: Minimum batch size + batch max wait time


Ehud-Lev-Forter:
Hi Team, Does it make sense to support `minimumBatchSize` with `maxTimeToWait` in Parallel consumer? The idea is that I want to use the PC as a micro batch consumer, but since my consumers are faster than my stream I get 1 event most of the time, I do get larger batches but only when system is in lag. Seems that I am not the only one that consider that: [see](https://github.com/confluentinc/parallel-consumer/issues/551)
What do you think about this idea?
Is this something you don't want to support? or it is just not the focus right now?

astubbs:
Sounds like a worthy feature to have! :)

rkolesnev:
I am working through the PR review on it - but i will wait until @astubbs has a look as well.

I am wondering on a concept a little bit though - what benefit is in enforcing min batch size - is it purely slight optimization for processing function / downstream resource use?

Generally batching is self balancing - if downstream processing is faster than consumption from kafka / flow rate - batching is not adding any advantage from purely processing speed, once flow rate exceeds time taken to process individual messages - batching starts to kick in. If processing explicitly depends on messages being batched - probably best bet to add small KStream / KSQL job on incoming topic to aggregate messages into batches (windowed etc).

From implementation side - is minimum batch size adding any value over just Linger ? similar to Kafka producer behaviour - i.e. keep current batch size and add linger (i.e. timeout for waiting to fill the batch)? 
i think that would make it slightly more familiar and a bit easier to reason about / slightly less complex implementation.

Ehud-Lev-Forter:
Hi @rkolesnev can you elaborate how can I "keep current batch size and add linger (i.e. timeout for waiting to fill the batch)" I thought linger is only in the producer side? 

As for best bet adding KStream / KSQL, if I understand correctly it means adding another hop to kafka which is more expensive. 

I totally agree that adding min batch size makes the library more complex and might harm performance.

Ehud-Lev-Forter:
Hi @rkolesnev, 
As for a, yes it is best effort, similar to some behavior of kafka connect using `partition.duration.ms` if I remember correctly.  
As for b, It is only relevant when the parallel consumer is meant to write back to kafka, but in many cases IMO it is meant to write to something else, thus batching using producer is not an option. 
As for c, I think most of the interesting use cases are for Key or unordered order. 

[...]

PR #562, opened 2023-03-03
feature: Add min batch size support

Ehud-Lev-Forter:
**Add min batch size support**
We would like to support minimum batch size. Since we might have a time with no traffic and we don't want to wait forever for the events that are not satisfy minimum batch, we will also add `minBatchTimeoutInMillis`, if we passed this timeout, code will be as today.
[see issue](https://github.com/confluentinc/parallel-consumer/issues/560)

### Checklist

- [ ] Changelog

### DIFF
--- a/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumerOptions.java
+++ b/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumerOptions.java
@@ -1,7 +1,7 @@
 package io.confluent.parallelconsumer;
 
 /*-
- * Copyright (C) 2020-2022 Confluent, Inc.
+ * Copyright (C) 2020-2023 Confluent, Inc.
  */
 
 import io.confluent.parallelconsumer.internal.AbstractParallelEoSStreamProcessor;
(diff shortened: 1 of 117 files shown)


astubbs:
Title should be something likes:

feature: Add min batch size support

rkolesnev (changes requested):
In general logical flow looks fine to me.
I would really like to see more tests and integration tests for it though - with transactions, different ordering modes (Unordered / Key /Partition) and 'normal' non-reactor usage mode of PC - it has a potential to introduce edge case issues that additional tests could shake out.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumerOptions.java
```diff
@@ -393,6 +419,23 @@ public void validate() {
...
 
         transactionsValidation();
+        validateMinBatchParameters();
+    }
+
+    private void validateMinBatchParameters() {
+        if (isEnforceMinBatch()){
+            if (minBatchSize > batchSize)
+                throw new IllegalArgumentException(
+                        msg("minBatchSize can not by bigger than batchSize: {} > {}",
```
eddyv, line 429:
```suggestion
                        msg("minBatchSize cannot be bigger than batchSize: {} > {}",
```
typo

eddyv (commented):
Seems to break a test 

```
[ERROR] io.confluent.parallelconsumer.state.WorkManagerTest.resumesFromNextShard(ProcessingOrder)[1] -- Time elapsed: 0.007 s <<< FAILURE!
14:36:51  expected: 1
14:36:51  but was : 2
```

Issue #546, opened 2023-02-09
Truncating state


lennehendrickx:
Hi @astubbs ,

We started using the parallel consumer recently for some of our services. Thanks for creating this great new consumer.
We sometimes see the following warnings in our logs and wonder what could be causing them.

We do not use compact topics and the retention is set to:
`retention.ms=604800000` (7 days)

The consumer offset reset is set to:
`auto.offset.reset=latest`

```
Truncating state - removing records lower than 208437. Offsets have been removed from the partition by the broker or committed offset has been raised. Bootstrap polled 208437 but expected 1 from loaded commit data. Could be caused by record retention or compaction and offset reset policy LATEST.
```
The expected value is also sometimes 0 instead of 1

We also see the following warning, which might be related
```
Bootstrap polled offset has been reset to an earlier offset (1081798) - truncating state - all records above (including this) will be replayed. Was expecting 1081799 but bootstrap poll was 1081798. Could be caused by record retention or compaction and offset reset policy EARLIEST.
```

I have also created a [PR](https://github.com/confluentinc/parallel-consumer/pull/545) to add the partition and topic to the truncate logging. 

Do you know how we could explain these warnings?

rkolesnev:
Hi, 

I have found a bug in offset encoding / decoding that can lead to those warnings in certain conditions.

The warning is only shown on partition assignment - and first poll - subsequent polls dont trigger this logic.
The logic is to load commit data (including any metadata about incomplete offsets below highest succeeded) and subsequently compare offset read from commit data and offset of first polled record - the warning is shown is record's offset is above the expected offset. 

The bug that i have found is in offset encoding / decoding when RunLength encoding is used and there are no incomplete offsets below highest seen / polled offset when partition is revoked and offsets are committed.
On decoding such offset metadata - highest succeeded offset is incorrectly returned as 1 (actually is decoded as 0 - but its calculated as last committed + 1).

The warning then reads Truncating state .... Bootstrap polled [OFFSET] but expected 1 from loaded commit data....

I am not sure why the other warning could be shown - that particular bug wouldn't cause that - Truncating state ... Was expecting 1081799 but bootstrap poll was 1081798.... 

I will raise a PR to fix that bug shortly and will incorporate the PR raised for the logging change - to include the Topic and Partition - into it.

lennehendrickx:
Hi @rkolesnev ,

Thanks for looking into this. With regards to the warnings I see that the following PR was included in the latest release:
[...]

PR #563, opened 2023-03-03
fix: runlength decoding after no-progress commit

rkolesnev:
When offset data is serialized using RunLength encoder - and base offset to commit is same as the next expected offset - 1 (or lower) - basically there is no extended metadata to encode - could be caused by slow processing (slower than consumption) or on rebalance etc.

When that data is subsequently decoded - the HighestOffset is returned as 0 - that causes the warning about state truncation to be shown and the bootstrap state truncation logic to be invoked.

This fix addresses the decoding side - initializing the HighestOffset = baseOffset-1 - same as bitset decoder, but potentially we could detect and skip encoding of metadata in those scenarios - need to dig deeper and can be done as follow up.

Related issue #546 

### Checklist

- [x] Changelog

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -14,6 +14,12 @@ ifndef::github_name[]
 toc::[]
 endif::[]
 
+== 0.5.2.6
+
+=== Fixes
+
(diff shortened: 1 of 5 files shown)


nachomdo (approved):
LGTM 👍

PR #594, opened 2023-06-28
Feature/health metrics micrometer

rkolesnev:
Micrometer based metrics
see #71 and #464 for background
### Checklist

- [X] Documentation
- [X] Changelog
- [X] Tests


parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/BrokerPollSystem.java
```diff
@@ -77,6 +83,13 @@ public BrokerPollSystem(ConsumerManager<K, V> consumerMgr, WorkManager<K, V> wm,
                 committer = Optional.of(consumerCommitter);
             }
         }
+        initMetrics();
+    }
+
+    private void initMetrics() {
+        statusGauge = PCMetrics.getInstance().gaugeFromMetricDef(PCMetricsDef.PC_STATUS, this, poller -> poller.runState.ordinal(), Tag.of("status", runState.name()));
```
eddyv, line 90: Wouldn't this create a separate metric per status? I am uncertain if this is ideal

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/WorkManager.java
```diff
@@ -299,19 +292,42 @@ public boolean isDirty() {
...
+            if (!succeededRecordsCounters.containsKey(topicPartition)) {
+                succeededRecordsCounters.put(topicPartition, PCMetrics.getInstance().getCounterFromMetricDef(PCMetricsDef.PROCESSED_RECORDS, getWorkManagerCounterTags(topicPartition)));
+            }
+            if (!failedRecordsCounters.containsKey(topicPartition)) {
+                failedRecordsCounters.put(topicPartition, PCMetrics.getInstance().getCounterFromMetricDef(PCMetricsDef.FAILED_RECORDS, getWorkManagerCounterTags(topicPartition)));
+            }
+        });
+    }
+
+    private void safelyIncrementCounter(Map<TopicPartition, Counter> counterMap, TopicPartition topicPartition) {
```
eddyv, line 313: Not crazy about this name. Just a nitpick

```suggestion
    private void incrementCounterIfPresent(Map<TopicPartition, Counter> counterMap, TopicPartition topicPartition) {
```

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/WorkManager.java
```diff
@@ -299,19 +292,42 @@ public boolean isDirty() {
...
 
-    private PCMetrics.WorkManagerMetrics getMetrics() {
-        return PCMetrics.WorkManagerMetrics.builder().
-        waitingRecords(getNumberOfWorkQueuedInShardsAwaitingSelection())
-                .inflightRecords(getNumberRecordsOutForProcessing()).build();
+    private Tag[] getWorkManagerCounterTags(TopicPartition topicPartition) {
+        return new Tag[]{Tag.of("topic", topicPartition.topic()), Tag.of("partition", String.valueOf(topicPartition.partition()))};
+    }
+
+    private void deregisterCounterMetrics(Collection<TopicPartition> partitions) {
```
eddyv, line 321: Should we make this more generic like `deregisterTopicPartitionMetrics`? There could be other types in the future. As an aside -- do we need to deregister those gauges here too?

eddyv (approved):
Nitpicks

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/metrics/PCMetrics.java
```diff
@@ -0,0 +1,200 @@
...
+    /**
+     * Meter registry used for metrics - set through init call on singleton initialization. Configurable through
+     * Parallel Consumer Options.
+     */
+    private MeterRegistry meterRegistry;
+
+    /**
+     * Tracking of registered meters for removal from registry on shutdown.
+     */
+    private List<Meter.Id> registeredMeters = new ArrayList<>();
```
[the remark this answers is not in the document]
rkolesnev: I am not sure i would want to clear whole registry - as it may be used for other things as well.
For example some logic like:
App managing PC instance (maybe even more than 1) and on whatever condition restarting 1 instance. Metrics registry would be shared across the app itself, any JVM metric collectors etc and PC instances. 
In that case on shut down of PC instance it should only clear the meters that it created and not all meters on registry - at least that is my thinking here.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/BrokerPollSystem.java
rkolesnev: Hmm, you are right, will test this and see if it needs to clear previous gauge on status change...

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/WorkManager.java
rkolesnev: Yep will rename.
Gauges dont need to be deregistered - only topic / partition specific metrics need created / deregistered - as topic / partition is assigned / revoked. 
Gauges are tracking overall count of inflight / waiting messages on WorkManager level so they are fine.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/metrics/PCMetrics.java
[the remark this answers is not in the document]
eddyv: Ah right makes sense.

eddyv (approved):
lgtm

Issue #559, opened 2023-02-23
Don't drain mode: worker threads are interrupted on shutdown


dannpopescu:
Hi,

I’ve already written about a bug we’ve discovered with the `DRAIN` shutdown mode here: https://github.com/confluentinc/parallel-consumer/issues/552. After we realized the PC keeps polling messages in the `DRAIN` mode we chose the `DONT_DRAIN` mode to gracefully shutdown the PC, but it behaves unexpectedly too.

The version in use is `0.5.2.4`.

Our use case is pretty standard I would say. We use only the core module and message processing involves network calls to other services. When calling them, errors are expected, so we commit the offsets in both the happy case and not-so-happy case.

The problem comes at the shutdown. In the `DONT_DRAIN` mode the PC will shutdown the worker threads via interrupts. This means that all in-progress messages will fail on network calls. Of course, we don’t want to commit the failed messages in this case. At the moment, to avoid committing the message’s offset, we catch all the exceptions thrown by the processing code and check if the root cause is an `InterruptedException`. If it’s the case then throw the exception out of the user function to the PC. This way the PC puts the message in the retry queue and doesn’t commit it. This approach works at the moment, but it’s kinda hacky I would say.

All this said, do you think it’s reasonable to change the way in which the worker thread pool is shutdown so that it doesn’t interrupt the worker threads, but instead lets the already scheduled tasks to finish gracefully? This is basically adhering to `DONT_DRAIN`'s current description:
> Stop downloading more messages, and stop procesing more messages in the queue, but finish processing messages already being processed locally.

antonmos:
@niamhthornbury I noticed that you closed https://github.com/confluentinc/parallel-consumer/pull/593  that trying to address this issue. Could you share any plans to address this?

PR #615, opened 2023-08-02
PL-176 handle close dont drain mode gracefully

rkolesnev:
Handle close dont drain mode gracefully.
Addresses issue - #559 

Added ParallelConsumerOptions shutdownTimeout - shutdown timeout is used during closeDontDrain shutdown - but because that is invoked on error as well - this option allows to set the timeout during PC creation.
Added ParallelConsumerOptions drainTimeout - to allow to specify time bound for draining the queues separately from shutdown timeout.
### Checklist

- [x] Changelog

### DIFF
--- a/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
+++ b/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
@@ -12,7 +12,6 @@ import io.confluent.parallelconsumer.metrics.PCMetricsDef;
 import io.confluent.parallelconsumer.state.WorkContainer;
 import io.confluent.parallelconsumer.state.WorkManager;
 import io.micrometer.core.instrument.Gauge;
-import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.binder.jvm.ExecutorServiceMetrics;
 import lombok.*;
 import lombok.extern.slf4j.Slf4j;
@@ -34,7 +33,6 @@ import java.time.Instant;
 import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
@@ -65,6 +63,14 @@ public abstract class AbstractParallelEoSStreamProcessor<K, V> implements Parall
      * user function.
      */
     private static final String MDC_WORK_CONTAINER_DESCRIPTOR = "offset";
+
+    /**
+     * Timeout used by various subsystems(BrokerPoller, Consumer) during shut down.
+     */
+    public static final Duration DEFAULT_TIMEOUT = Duration.ofSeconds(30);
+
+    public static final Duration GRACE_PERIOD_FOR_OVERALL_SHUTDOWN = Duration.ofSeconds(10);
+
 
     @Getter(PROTECTED)
     protected final ParallelConsumerOptions<K, V> options;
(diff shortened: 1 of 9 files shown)


parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -696,11 +730,15 @@ protected <R> void supervisorLoop(Function<PollContextInternal<K, V>, List<R>> u
                     controlLoop(userFunctionWrapped, callback);
                 } catch (InterruptedException e) {
                     log.debug("Control loop interrupted, closing");
-                    doClose(DrainingCloseable.DEFAULT_TIMEOUT);
+                    Thread.interrupted(); //clear interrupted flag as during close need to acquire commit locks and interrupted flag will cause it to throw another interrupted exception.
+                    doClose(shutdownTimeout);
                 } catch (Exception e) {
+                    if (Thread.interrupted()) { //clear interrupted flag
```
rkolesnev, line 736: @jhollandus, @eddyv - can you have a look at this - seems this control thread can be interrupted but we need it to not have interrupted flag set when doing doClose() as there is some code that is sensitive to that flag..

eddyv: From what I understand, since we are actually actively trying to perform some actions as a result of an interrupt, it seems fine to me to clear the flag to perform some work. We may want to consider re-throwing the flag after we are done with the close? See [stackoverflow question](https://stackoverflow.com/questions/10401947/methods-that-clear-the-thread-interrupt-flag)?

rkolesnev: Dont really see a point resetting the interrupt flag - the only way to reset it is to interrupt the thread again - and that will lead to InterruptedException (unless the thread is blocked etc). But it is a shutdown process - so after we are done with it - the PC - including that control-thread is shut down / closed.

rkolesnev: Yeah - the next thing after finishing doClose - the while(state!=CLOSED) loop will finish (as we transition to CLOSED state during the doClose call) and the Future will be set / done. So the control thread will be done at that point.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -66,6 +64,14 @@ public abstract class AbstractParallelEoSStreamProcessor<K, V> implements Parall
      */
     private static final String MDC_WORK_CONTAINER_DESCRIPTOR = "offset";
 
+    /**
+     * Timeout used by various subsystems(BrokerPoller, Consumer) during shut down.
+     */
+    public static final Duration DEFAULT_TIMEOUT = Duration.ofSeconds(30);
```
eddyv, line 70: Feels like something that should be loaded as a config? If it's shared in multiple classes it feels a bit weird to be controlled here. Should it be in `ParallelConsumerOptions`?

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
eddyv: Great

eddyv (approved):
Small QOL suggestion. Otherwise LGTM

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
rkolesnev: Not really sure - it was hard coded in the DrainMode enum but i've added config for shutdown and drain operations in the ParallelConsumerOptions. 
Still need this default timeout for other operations that were using it... i dont really want to make it configurable just yet as its used for commit response blocking timeout and for poller subsystem shutdown.

Issue #618, opened 2023-08-05
Metrics error with 0.5.2.6


makghub9:
Sometimes I'm getting below error with the newer version jar. It happens on startup only sometimes. If I remove the meter registry from the parallel consumer, this won't happen.

2023-08-05 17:44:24,380 [http-nio-8083-exec-2] Failed to apply the value function for the gauge 'pc.partitions.paused'. Note that subsequent logs will be logged at debug level.
java.util.ConcurrentModificationException: KafkaConsumer is not safe for multi-threaded access
	at org.apache.kafka.clients.consumer.KafkaConsumer.acquire(KafkaConsumer.java:2491)
	at org.apache.kafka.clients.consumer.KafkaConsumer.acquireAndEnsureOpen(KafkaConsumer.java:2475)
	at org.apache.kafka.clients.consumer.KafkaConsumer.paused(KafkaConsumer.java:2066)
	at io.confluent.parallelconsumer.internal.ConsumerManager.paused(ConsumerManager.java:133)
	at io.confluent.parallelconsumer.internal.BrokerPollSystem.lambda$initMetrics$1(BrokerPollSystem.java:92)
	at io.micrometer.core.instrument.StrongReferenceGaugeFunction.applyAsDouble(StrongReferenceGaugeFunction.java:48)
	at io.micrometer.core.instrument.internal.DefaultGauge.value(DefaultGauge.java:53)
	at io.micrometer.prometheus.PrometheusMeterRegistry.lambda$newGauge$5(PrometheusMeterRegistry.java:331)
	at io.micrometer.prometheus.MicrometerCollector.collect(MicrometerCollector.java:75)
	at io.prometheus.client.Collector.collect(Collector.java:45)
	at io.prometheus.client.CollectorRegistry$MetricFamilySamplesEnumeration.findNextElement(CollectorRegistry.java:204)
	at io.prometheus.client.CollectorRegistry$MetricFamilySamplesEnumeration.nextElement(CollectorRegistry.java:219)
	at io.prometheus.client.CollectorRegistry$MetricFamilySamplesEnumeration.nextElement(CollectorRegistry.java:152)
	at io.prometheus.client.exporter.common.TextFormat.writeOpenMetrics100(TextFormat.java:202)
	at org.springframework.boot.actuate.metrics.export.prometheus.TextOutputFormat$2.write(TextOutputFormat.java:62)
[...]

PR #620, opened 2023-08-06
fix: Return previous value for 'partitions.paused' on ConcurrentModificationException

acktsap:
Resolves: https://github.com/confluentinc/parallel-consumer/issues/618

### Checklist

- [ ] Changelog


youribonnaffe:
Isn't something we would like to handle in https://github.com/confluentinc/parallel-consumer/blob/master/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/ConsumerManager.java the same way metadata are cached between polls?

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/BrokerPollSystem.java
```diff
@@ -88,8 +91,21 @@ public BrokerPollSystem(ConsumerManager<K, V> consumerMgr, WorkManager<K, V> wm,
 
     private void initMetrics() {
         statusGauge = PCMetrics.getInstance().gaugeFromMetricDef(PCMetricsDef.PC_POLLER_STATUS, this, poller -> poller.runState.getValue());
+
+        AtomicInteger pausedPartitionHolder = new AtomicInteger();
```
jhollandus, line 95: I would suggest this be pulled out into a final class variable.

jhollandus (changes requested):

jhollandus:
> Isn't something we would like to handle in https://github.com/confluentinc/parallel-consumer/blob/master/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/ConsumerManager.java the same way metadata are cached between polls?

This is a good idea. It would be best to split the `ToDoubleFunction` away from the `AtomicInteger` and put the integer into the `ConsumerManager` poll loop.  Finally reference the pauses partitions from the consumer manager in the ToDoubleFunction.

jhollandus (changes requested):
See https://github.com/confluentinc/parallel-consumer/pull/620#issuecomment-1673801868

acktsap:
Hi, thanks for comments. I moved cache logic into `ConsumerManager`.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/ConsumerManager.java
```diff
@@ -34,6 +37,8 @@ public class ConsumerManager<K, V> {
      */
     private ConsumerGroupMetadata metaCache;
 
+    private final AtomicReference<Set<TopicPartition>> pausedPartitionsCache = new AtomicReference<>();
```
eddyv, line 40: I feel like we can just keep track of the size itself.
```suggestion
    @Getter
    private final AtomicInteger pausedPartitionSize = new AtomicInteger(0);
```

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/ConsumerManager.java
```diff
@@ -65,8 +70,9 @@ ConsumerRecords<K, V> poll(Duration requestedLongPollTimeout) {
         return records;
     }
 
-    protected void updateMetadataCache() {
+    protected void updateCache() {
         metaCache = consumer.groupMetadata();
+        pausedPartitionsCache.set(consumer.paused());
```
eddyv, line 75: If you accept the previous change.
```suggestion
        pausedPartitionSize.set(consumer.paused().size());
```

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/ConsumerManager.java
```diff
@@ -130,7 +136,7 @@ public void pause(final Set<TopicPartition> assignment) {
     }
 
     public Set<TopicPartition> paused() {
-        return consumer.paused();
+        return pausedPartitionsCache.get();
```
eddyv, line 139: revert
```suggestion
        return consumer.paused
```

eddyv (changes requested):
It looks like the latest changes failed a few tests.

I offered some suggestions so that we just keep track of the size itself, which seems to address this issue.

Will need to add the following in `BrokerPollSystem.initMetrics()` if you choose to go that route.

```java
-                this.consumerManager, consumerManager -> consumerManager.paused().size());
+                this.consumerManager, consumerManager -> consumerManager.getPausedPartitionSize().get());
```

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/ConsumerManager.java
acktsap: Done

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/ConsumerManager.java
acktsap: I reflected the review except for few changes.

- Use `volatile int` type and named `pausedPartitionSizeCache` for consistency
- Add getter named `get PausedPartitionSize()` returning `pausedPartitionSizeCache` value

eddyv (approved):
lgtm thanks 👍

jhollandus (approved):
LGTM

Issue #606, opened 2023-07-19
Parallel consumer stops processing data sometimes


bdeneuter:
Hi,

In production we noticed that the parallel consumer (0.5.2.5) sometimes stops processing data. It is a similar problem like https://github.com/confluentinc/parallel-consumer/issues/547.
We can not reproduce the problem but we noticed the following:

- the number of fetches goes to 0 or close to 0
- a restart fixes the problem

When looking into the fix for 547, I was wondering if the fix is complete?
The fix validates if a WorkContainer is stale and end the flight of the WorkContainer and changes `numberRecordsOutForProcessing`
https://github.com/confluentinc/parallel-consumer/blob/master/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/WorkManager.java#L251

Are those WorkContainers also cleaned up from the ProcessingShard which keeps a set of WorkContainers and is used to fetch work?
https://github.com/confluentinc/parallel-consumer/blob/master/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ProcessingShard.java#L44

For the succes and failure case, the ShardManager is being called while for the stale case the ShardManager is not called:
https://github.com/confluentinc/parallel-consumer/blob/master/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/WorkManager.java#L152
https://github.com/confluentinc/parallel-consumer/blob/master/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/WorkManager.java#L173

Kind regards,
Bart

johnbyrnejb:
We will investigate but we are planning to release a new version this week which will include metrics. Can you upgrade your version at that point and see if you can capture any more information that may be useful to triage? 

Thanks

bdeneuter:
Looking forward for the new version. Metrics would definitely help us to investigate this.

krvajal:
I am experiencing the same issue. Looking forward to the metrics update

acktsap:
Hi, i also experiencing the same issue..

sangreal:
I also experience this issue and after some digging I find the root cause, I am trying to fix this on my end, maybe will create a PR for this later.

10000-ki:
Hi, i also experiencing the same issue too..

eddyv:
Closed by #623

PR #623, opened 2023-08-08
fix : for issue 606 : exclude stale containers from counting and add remove add remove staled work containers function in ProcessingShard

sangreal:
Description...
This is the fix for https://github.com/confluentinc/parallel-consumer/issues/606

The issue happens when rebalancing and partitions got re-assigned:
1. The `workcontainer` will be marked as stale container since its epoch is mismatch between `partitionsAssignmentEpoch` of `PartitionState` and `epoch` from `WorkContainer`
https://github.com/confluentinc/parallel-consumer/blob/master/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/PartitionState.java#L655

2. `inFlight` of `WorkContainer` will be stay false since they will be filtered to be selected in `runUserFunction` in `AbstractParallelEoSStreamProcessor`
https://github.com/confluentinc/parallel-consumer/blob/master/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java#L1256

3. The consumer paused will be triggered since if the number of staled messages are big enough `isSufficientlyLoaded`  in `WorkManager` will return true.
https://github.com/confluentinc/parallel-consumer/blob/master/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/WorkManager.java#L232C1-L232C1
https://github.com/confluentinc/parallel-consumer/blob/master/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/WorkContainer.java#L230

4. Since the staled `workcontainer` can never be selected to run user function. The consumer will be paused infinitely until next-time rebalancing
https://github.com/confluentinc/parallel-consumer/blob/master/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/BrokerPollSystem.java#L304

5. Meanwhile, while `getWorkIfAvailable` in `WorkManager` will try to extract active workcontainers, but since the staled containers have the lower offsets, they will be extracted always earlier than the new ones. (if the same partitions have been assigned to the consumer again). Then the new records have no chances to be processed since the it will break out of the while loop.
https://github.com/confluentinc/parallel-consumer/blob/master/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ProcessingShard.java#L134

The proposed solution:
1. We should exclude those stale work containers from calculation
2. remove staled work containers from shard when partition assigned and revoke happen

### Checklist

- [x] Changelog

### DIFF
--- a/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ProcessingShard.java
+++ b/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ProcessingShard.java
@@ -80,7 +80,7 @@ public class ProcessingShard<K, V> {
         return entries.values().stream()
                 // todo missing pm.isBlocked(topicPartition) ?
                 // exclude stale workContainers from counting
-                .filter(workContainer -> !pm.getPartitionState(workContainer).checkIfWorkIsStale(workContainer))
+                .filter(workContainer -> !isWorkContainerStale(workContainer))
                 .filter(WorkContainer::isAvailableToTakeAsWork)
                 .count();
     }
@@ -105,13 +105,10 @@ public class ProcessingShard<K, V> {
         var slowWork = new HashSet<WorkContainer<?, ?>>();
         var workTaken = new ArrayList<WorkContainer<K, V>>();
 
-        if (removeStaledWorkContainersFromShard()) {
-            log.trace("have staled work containers removed");
-        }
-
         var iterator = entries.entrySet().iterator();
         while (workTaken.size() < workToGetDelta && iterator.hasNext()) {
             var workContainer = iterator.next().getValue();
+
 
             if (pm.couldBeTakenAsWork(workContainer)) {
                 if (workContainer.isAvailableToTakeAsWork()) {
@@ -129,6 +126,11 @@ public class ProcessingShard<K, V> {
                     log.trace("Processing by {}, so have cannot get more messages on this ({}) shardEntry.", this.options.getOrdering(), getKey());
                     break;
                 }
+            } else if (isWorkContainerStale(workContainer)) {
+                // stale work containers are handled here
+                // the normal work containers' offset should be higher than stale work containers
+                // the break below should be happened after all stale work containers
+                log.trace("the work container {} is stale, continue scanning for the next", workContainer);
             } else {
                 // break, assuming all work in this shard, is for the same ShardKey, which is always on the same
                 //  partition (regardless of ordering mode - KEY, PARTITION or UNORDERED (which is parallel PARTITIONs)),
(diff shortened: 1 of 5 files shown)


parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ProcessingShard.java
```diff
@@ -182,4 +188,15 @@ private static String cantTakeAsWorkMsg(WorkContainer<?, ?> workContainer, Durat
...
     }
+
+    // remove staled WorkContainer otherwise when the partition is reassigned, the staled messages will:
+    // 1. block the new work containers to be picked and processed
+    // 2. will cause the consumer to paused consuming new messages indefinitely
+    private boolean removeStaledWorkContainersFromShard() {
+        return this.entries.entrySet()
+                .removeIf(entry -> {
+                    WorkContainer<K, V> workContainer = entry.getValue();
+                    return pm.getPartitionState(workContainer).checkIfWorkIsStale(workContainer);
```
eddyv, line 199: This call here kills performance (see the test `lessKeysThanThreads` in `ParallelEoSStreamProcessorTest`). I wonder if there's a way for us to reduce the number of calls when cleaning up workers. We already perform the filtering, so we might want a separate job to handle stale worker removal.

@jhollandus

sangreal: Thanks for pointing out for I miss the integration test. Let me think about this one.

sangreal: @eddyv please help check the updates. I think adding another thread/job for removal of stale worker could be an overkill.
So I filter them out without break out of the loop to let normal workers could be picked up.

eddyv: Hmm... still seems to be a bit on the slow side even for just filtering. Better but still slow. I suggested the use of an extra thread, purely just so we can reduce the number of filter calls we make. Essentially, have a background thread that cleans things up every few seconds or something. Open to suggestions

sangreal: @eddyv Thanks for the comments. I put the removal of stale containers into `onPartitionsAssigned` and `onPartitionsRevoked` since:
1. these two places are the only places that the containers will be marked as stale.
2. the extra thread may introduce extra context switch overhead.

sangreal: @eddyv @jhollandus, just a kind reminder for checking the updated PR

eddyv: Thanks -- sorry, life gets in the way sometimes.

It looks like the build still times out. (btw, you can actually view the logs yourself by navigating to public ci check and clicking on view more details)

image

At first glance, it could be that making this call `.filter(workContainer -> !isWorkContainerStale(workContainer))` is expensive?

I wonder if it's worth it to just exclude this filter and rely on `onPartitionsAssigned` and `onPartitionsRevoked`.

sangreal: @eddyv Thanks for your valuable advice. I totally agree with you. The removal in `onPartitionsAssigned` and `onPartitionsRevoked` should be enough.
I have removed the filter part and the tests in Jenkins is passed.

Thanks for the tips for check the CI!

eddyv (approved):
LGTM

sangreal:
@eddyv Thanks a lot for your quick response. Actually we are kind of blocked by this issue, when could we expect this fix merged?
And when could we have a snapshot version or release version to use?

eddyv:
Thank you!

sangreal:
@eddyv  Thank you so much for your help here! We are rolling out parallel consumer to company wide usage, this fix is very crucial for us.
Thanks again and looking forward to more interaction in the future.

PR #624, opened 2023-08-10
fix readme doc

flashmouse:
In ReadMe. it introduce the metric function.
```
    var options = ParallelConsumerOptions.<String, String>builder()
            .ordering(ParallelConsumerOptions.ProcessingOrder.KEY)
            .maxConcurrency(1000)
            .consumer(kafkaConsumer)
            .meterRegistry(meterRegistry)                     //(1)
            .metricsTags(Tags.of(Tag.of("instance", "pc1")))    //(2)
            .build();

......
    kafkaClientMetrics = new KafkaClientMetrics(kafkaConsumer); //(3)
    kafkaClientMetrics.bindTo(meterRegistry);                 //(4)
```
with gloss
```
1. Optional - Kafka Consumer Micrometer metrics object created for Kafka Consumer that is later used for Parallel Consumer.

2. Optional - Kafka Consumer Micrometer metrics are bound to Meter Registry.

3. Meter Registry is set through ParallelConsumerOptions.builder(), if not specified - will default to CompositeMeterRegistry - which is No-op.

4. Optional - "instance" tag with value of "pc1" is set through same builder - it will be added to all Parallel Consumer meters
```
I think the actual sequence should be 3,4,1,2

i'm not sure whether should modify this file, i cannot execute ``mvn asciidoc-template::build`` in my own pc.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1211,10 +1211,10 @@ Following example illustrates setup of Parallel Consumer with Meter Registry and
         return eosStreamProcessor;
     }
 ----
-<1> - Optional - Kafka Consumer Micrometer metrics object created for Kafka Consumer that is later used for Parallel Consumer.
-<2> - Optional - Kafka Consumer Micrometer metrics are bound to Meter Registry.
-<3> - Meter Registry is set through ParallelConsumerOptions.builder(), if not specified - will default to CompositeMeterRegistry - which is No-op.
-<4> - Optional - "instance" tag with value of "pc1" is set through same builder - it will be added to all Parallel Consumer meters
+<1> - Meter Registry is set through ParallelConsumerOptions.builder(), if not specified - will default to CompositeMeterRegistry - which is No-op.
+<2> - Optional - "instance" tag with value of "pc1" is set through same builder - it will be added to all Parallel Consumer meters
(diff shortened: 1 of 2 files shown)


rkolesnev (changes requested):
Yep, good catch.
Can you please regenerate README as well - so that its updated in this PR too?

rkolesnev (approved):
Will regenerate README on release prep.

flashmouse:
it's hard for me, I could see <templateFile>README_TEMPLATE.adoc</templateFile> in pom.xml but after run ``asciidoc-template:build `` it throws: ``java.nio.file.NoSuchFileException: src/docs/README-template.adoc``, seems this plugin modified the value of ``templateFile``

rkolesnev:
> it's hard for me, I could see `<templateFile>README_TEMPLATE.adoc</templateFile>` in `pom.xml` but after run `asciidoc-template:build ` it throws: `java.nio.file.NoSuchFileException: src/docs/README-template.adoc`, seems this plugin modified the value of `templateFile`

Yeah - just noticed as well - it still generates the updated README. it looks like the error is due to the plugin trying to execute not just for parent project but for sub-modules as well - where template path is not defined. 
I've pushed updated README - no worries.

PR #628, opened 2023-08-25
PL-464 remove micrometer-atlas dependency

rkolesnev:
Remove micrometer-atlas dependency. 
It is causing issues in Spring as it autoconfigures use of Atlas registry which may not be present and causes attempts to send metrics that fail with warnings.

It was added by mistake and not actually needed. 
### Checklist

- [X] Changelog

### DIFF
--- a/parallel-consumer-core/pom.xml
+++ b/parallel-consumer-core/pom.xml
@@ -40,10 +40,6 @@
             <artifactId>micrometer-core</artifactId>
             <scope>compile</scope>
         </dependency>
-        <dependency>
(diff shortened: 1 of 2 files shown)


eddyv (approved):

krvajal:
@rkolesnev any idea when the next release will be available?

rkolesnev:
> @rkolesnev any idea when the next release will be available?

@krvajal - can't give a specific date, but we should have it released soon - mainly a bugfix realease.

Issue #637, opened 2023-09-05
Parrallel consumer stuck in infinite rebalance


bmaggi:
Hi,

I have a quite complexe issue with the 2 latest versions 0.5.2.6 and 0.5.2.7 while everything works correctly in 0.5.2.5. 

When the cpu increase we add machine to distribute the load between them. This obviously trigger a rebalance.
With 0.5.2.6 or 0.5.2.7, the rebalance is so long (> 5 min) that something seems to be blocking a correct rebalance.

I don't see anything particularly meaningful in kafka consumer logs

But I see a lot of _"isPoolQueueLow(): Max loading factor steps reached: 100/100"_ and _"Error committing offsets : Offset commit cannot be completed since the consumer is undergoing a rebalance for auto partition assignment. You can try completing the rebalance by calling poll() and then retry the operation."_ which are quite expected since the rebalance is stuck. 

Fixing the number of machine and restarting solve the issue so I was expecting the fix of https://github.com/confluentinc/parallel-consumer/issues/606 to solve this but unfortunally this was not the case. 

I don't now which logs or metrics may be of use to solve or explain this. 
Feel free to ask for more informations.

Regards
Benoît

sangreal:
I am interesting in the issue, could u please share the trace log of parallel consumer?

bmaggi:
Hi,

Here are the logs of parallel-consumer version 0.5.2.7 in case of 
* a bump of 3 hosts creating a rebalance. 
* rebalance stuck 
* decrease of 3 hosts 

I removed the warn logs, huge spam of "_isPoolQueueLow(): Max loading factor steps reached: 100/100_" which means to me that parallel-consumer is stuck. 

And all _Error committing offsets_ are related to _Offset commit cannot be completed since the consumer is undergoing a rebalance for auto partition assignment. You can try completing the rebalance by calling poll() and then retry the operation._

[extract-parallel-consumer-0527.csv](https://github.com/confluentinc/parallel-consumer/files/12557574/extract-parallel-consumer-0527.csv)

sangreal:
#606 is to solve the issue for reassigning the same partition to the same instance without revoking.
Your case is probably another new issue. Do you have trace level log for parallel consumer turned on?

colinkuo:
This could be related to #402

rkolesnev:
Hi @bmaggi, 

[...]

PR #665, opened 2023-11-17
Fix synchronisation logic between commits

rkolesnev:
Fix synchronisation logic between transactional producer commit and partition revocation commit

* Fixes https://github.com/confluentinc/parallel-consumer/issues/637

### Checklist

- [X] Documentation (if applicable)
- [X] Changelog

### DIFF
--- a/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
+++ b/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
@@ -408,21 +408,13 @@ public abstract class AbstractParallelEoSStreamProcessor<K, V> implements Parall
     public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
         log.debug("Partitions revoked {}, state: {}", partitions, state);
         isRebalanceInProgress.set(true);
-        while (this.producerManager.map(ProducerManager::isTransactionCommittingInProgress).orElse(false))
+        while (isTransactionCommittingInProgress())
             Thread.sleep(100); //wait for the transaction to finish committing
 
         numberOfAssignedPartitions = numberOfAssignedPartitions - partitions.size();
 
         try {
             // commit any offsets from revoked partitions BEFORE truncation
-            this.producerManager.ifPresent(pm -> {
-                try {
-                    pm.preAcquireOffsetsToCommit();
-                } catch (Exception exc) {
-                    throw new InternalRuntimeException(exc);
-                }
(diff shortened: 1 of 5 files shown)


parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -408,21 +408,13 @@ public void subscribe(Pattern pattern, ConsumerRebalanceListener callback) {
...
         try {
             // commit any offsets from revoked partitions BEFORE truncation
-            this.producerManager.ifPresent(pm -> {
-                try {
-                    pm.preAcquireOffsetsToCommit();
-                } catch (Exception exc) {
-                    throw new InternalRuntimeException(exc);
-                }
-            });
-
```
eddyv, line 425: I'm guessing this was the culprit?

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -1430,6 +1422,11 @@ private void clearCommitCommand() {
         }
     }
 
+    private boolean isTransactionCommittingInProgress() {
+        return options.isUsingTransactionCommitMode() &&
```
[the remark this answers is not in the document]
rkolesnev: it should be, but i am just playing it safe - IMHO - i prefer to explicitly check producer transaction state only if transactional producer is actually specified in options rather than check its state regardless and rely on default return to be false for non-transactional producer.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
rkolesnev: yes, kind of - on its own - its not a big deal to have it here - but because it is done without check for transactional producer option - this was called even when non-transactional producer was specified - but the `pm.preAcquireOffsetsToCommit();` is only specifically for Transactional producer handling... so it was causing issue by setting write lock which was not released as committer was not transactional producer one...

rkolesnev: but it is not really required as `commitOffsetsThatAreReady();` few lines of code down already calls `preAcquireOffsetsToCommit()` as part of commit process.

PR #667, opened 2023-11-19
add multiple caches for accelerating available container count calculation

sangreal:
Description...
This pr is for reopening for previous approval pr : https://github.com/confluentinc/parallel-consumer/pull/644

### Explanation
1. introduce cache for ready-to-be-retried container count (could be selected to retry containers). Therefore the 
**actually available count = all available container count - (still pending for running retry containers count)**
2. The caches calculation happens when containers are selected to run user function (before / after)
3. The available work containers will change only after the container selection, therefore it is accurate to update the cache at this point.
4. no extra threading needed in this solution comparing to previous one
5. The improvement is substantial. The performance of lessKeysThanThreads  has been improvement from 01:24 -> 00:38

### prev flow
cache_improvement_prev drawio

### current flow
cache_improvement_nowdrawio drawio

### Possible Question
Q : Previously the expired retry is calculated on fly which looks more accurate than new flow?
A : Actually they will be eventually same since for `controlLoop`, it will wait for <= (latest retry time) and it will update the caches. Previous flow, the available work container count also only update after (latest retry time)

### Checklist

- [x] Changelog


eddyv (dismissed):
lgtm. previous PR for reference: https://github.com/confluentinc/parallel-consumer/pull/644

rkolesnev (changes requested):
In principle the idea looks good, but i do not really like that we are reserving / busy spinning a thread to implement this.
Additionally - i think there may be a flaw in the logic of how retry poll / available work is marked based on retry handler... happy to discuss here or on confluent community slack if it makes sense..
If i delay retryQueue.poll a little (through debugger) to trigger race with picking work - then the counter is off - so it really depends on updates in correct order based on timing of things...

Will create a branch illustrating this shortly.

sangreal:
@rkolesnev Thanks for your comments. I am open for your advices if you have better idea.
But I want to point out that the performance gain is substantial and I kind of interesting to know how the counter will be off.
And I think I also agree with you that less threading model will be always a better choice.
Anyway, looking forward to your example. We could also chat in the slack

sangreal:
@rkolesnev @eddyv could you help review again?
I have introduced another way of implementation without extra threading to address @rkolesnev 's concern.

rkolesnev:
> @rkolesnev @eddyv could you help review again? I have introduced another way of implementation without extra threading to address @rkolesnev 's concern.

@sangreal - yes I had a quick look - will review / step through in depth when I have a chance, thanks!

sangreal:
@rkolesnev Thanks for letting me know the jenkins is recovered. I have updated pr for minor issues to fix the jenkins build.
Please help review when you have time.
Thanks

rkolesnev:
There is a bit missing for tracking the counts in [ProcessingShard.removeStaleWorkContainersFromShard()](https://github.com/confluentinc/parallel-consumer/blob/de3805555f7cfa99e4b6ebb29c8859b424cf819f/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ProcessingShard.java#L10) - i think it would need to remove from retryQueue, check and decrement expiredRetryContainerCnt and the main availableWorkContainerCnt.

rkolesnev:
Hmm the more i am getting my head round this the more i think we should solve it in two-fold approach. 
Basically we have 3 conditions that make up that count - container meets:
`isNotInFlight() && !isUserFunctionSucceeded() && isDelayPassed()`
We can directly track the `isNotInFlight` and `!isUserFunctionSucceeded` through the counter - as there are specific events that trigger their change - taken as work, succeeded, failed, etc.
But the last one - `isDelayPassed()` is time based - so we cannot really directly track its state from events - as there are no events - we really should be just checking that by observing during the call to `getNumberOfWorkQueuedInShardsAwaitingSelection().`

So what i am proposing is to keep the `availableWorkContainerCnt` and use it to track addition / removal of work containers from shard, and track the `isNotInFlight` and `!isUserFunctionSucceeded()` - which it basically does now in this PR, but at the same time remove `expiredRetryContainerCnt` and `retryItemCnt` and still scan the retryQueue - at least up to the first container that has its delay not yet expired (as its ordered by retry delay).

sangreal:
@rkolesnev Thanks a lot for your review. I think your solution also a good thought but still have to scan through the retryQueue.

I totally understand your concern but if you could take a look at the above flow diagram and some explanations.
The updates in `getNumberOfWorkQueuedInShardsAwaitingSelection` could be covered in `controlLoop`. Because it will only wait for <= (latest retry time = first retryable time) and it will update all caches. 
In previous flow, the available work container count also only update after (latest retry time)

https://github.com/confluentinc/parallel-consumer/blob/master/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java#L771

Let me give you an example:
Say now is 00:00, there are two retry items to be retried at 00:02 & 00:05. and there are no pending containers 
### for old flow:
1. it will check during every polling in BrokerPollSystem, but the count will stay same which is 0 until 00:02, the count will be 1 and after it processed, it will be back to zero. Until 00:05, the second one will be picked up, then the count will be 1 again.
### for new flow:
1. during every polling in BrokerPollSystem, the count will be not changed since cache is unchanged.
2. in `controlLoop` , the `Duration timeToBlockFor = shouldTryCommitNow ? Duration.ZERO : getTimeToBlockFor();` , therefore the `timeToBlockFor` will be <= min(all retry container waiting time). So at 00:02, the controlLoop will be run and updates (there is no blocking call in this loop), and count (`expiredRetryContainerCnt` -> 1) will be update to 1. Same applied to the second until 00:05, `controlLoop` will be processed to update the count to 1. 
### Conclusion
1. so as you could see, the updates timing is the same between previous and new flow

rkolesnev:
The problem with updating the available work that is in retry queue and has its delay elapsed is that the `expiredRetryContainerCnt` is only ever increased in `getWorkIfAvailable()` - and that only scans up to the number of work requested `workToGetDelta`:
```
         while (workTaken.size() < workToGetDelta && iterator.hasNext()) {
            var workContainer = iterator.next().getValue();

            if (pm.couldBeTakenAsWork(workContainer)) {
                if (workContainer.isAvailableToTakeAsWork()) {
                    log.trace("Taking {} as work", workContainer);

                    // only increase the ExpiredRetryContainerCnt when this is retry due since already added in the new container creation
                    if (workContainer.isDelayExistsExpired()) {
                        expiredRetryContainerCnt.incrementAndGet();
                    }
                    workContainer.onQueueingForExecution();
                    workTaken.add(workContainer);
                } else {
                ...
```
For example if we have 1000 items in the retry queue, all with delay of 5 seconds - once the delay elapsed - all 1000 are available for work - but we would only update the available work as we are taking them into processing and say with 16 processing threads - we would take only 16 and the rest will still be in retry queue but not counted as available work.
On top of that - as we are taking them as work - they will get marked as inFlight and be excluded from the available work count straight away. 
So effectively they will never show up as available work.
Implications of this is that BrokerPoller will be polling for more work instead of applying back-pressure and pausing consumers - which may lead to OOM / really deep queues etc in extreme / edge cases.

It all goes back to the fact that the retry delay is time based - and the only way to really know which/how many containers in retry queue have their retry delay elapsed and are available as work - is to scan the retry queue.

It does not have to be a full scan - as the retry queue is sorted by `nextRetryDueAt` - so it would be enough to scan up to the first container that still has the delay that is not elapsed to get the count of available / not available as work in retry queue.

In general i don't think scanning retry queue will have big performance implications - with thinking that - if the retry queue is small - then the scan is fast and if the retry queue is large - than we are in bad state anyway and probably not that concerned about the overhead introduced by scanning it - as processing is already slowed down by having a lot of messages to retry.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/WorkContainer.java
```diff
@@ -109,6 +109,18 @@ public boolean isDelayPassed() {
         return negative;
     }
 
+    /**
+     *
+     * @return whether this is retry and should be executed
+     */
+    public boolean isDelayExistsExpired() {
```
sangreal, line 116: I will leave it here since this is util method in case of future usage

sangreal:
@rkolesnev I have updated the pr according to your suggestions, meanwhile I keep the `retryItemCnt` since this count is accurate and will help the accelerate the expired items calculation while `expiredRetryContainerCnt` is removed.
Please help review again.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ShardManager.java
```diff
@@ -92,6 +96,9 @@ public class ShardManager<K, V> {
     @Getter(AccessLevel.PACKAGE) // visible for testing
     private final NavigableSet<WorkContainer<?, ?>> retryQueue = new TreeSet<>(retryQueueWorkContainerComparator);
 
+    @Getter
+    private final AtomicLong retryItemCnt = new AtomicLong(0);
```
rkolesnev, line 100: Any benefit to use `retryItemCnt` over the `retryQueue.size()` ? 
retryQueue is a TreeSet backed by TreeMap - which has `size` variable - that it keeps updated on changes to the TreeMap - so it should be similarly performant as its a cached value and doesn't require scan of the queue.

In addition removal of  `retryItemCnt` and using `retryQueue.size()` directly means code tracking changes to `retryQueue` to update the `retryItemCnt` is unnecessary - which is a plus.
For example below:
```
 if (this.retryQueue.remove(removedWC)) {
                retryItemCnt.decrementAndGet();
            }
 ```
can be reverted back to just 
```
this.retryQueue.remove(removedWC);
```

File: parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ShardManager.java [hunk 3]
```diff
@@ -298,4 +321,18 @@ private void initMetrics() {
         numberOfShardsGauge = pcMetrics.gaugeFromMetricDef(PCMetricsDef.NUMBER_OF_SHARDS,
                 this, shardManager -> shardManager.processingShards.keySet().size());
     }
+
+    // get expired items count from retryQueue
+    private long getExpiredItemCntFromRetryQueue() {
```
rkolesnev, line 326: maybe better name?
```suggestion
    private long getNumberOfFailedWorkReadyToBeRetried() {
```

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ShardManager.java
```diff
@@ -298,4 +321,18 @@ private void initMetrics() {
...
+        long count = 0;
+        for (WorkContainer<?, ?> workContainer : retryQueue) {
+            if (workContainer.getDelayUntilRetryDue().isNegative() || workContainer.getDelayUntilRetryDue().isZero()) {
+                count ++;
```
rkolesnev, line 330: should do another check before incrementing count as WorkContainer can be still in retryQueue but at the same time already picked for processing - so should be using either `workContainer.isAvailableToTakeAsWork()` or 
`isNotInFlight() && !isUserFunctionSucceeded()` as we already verified that delay passed by this point.
so basically it would be 
```
// count work that is not waiting for retry backoff delay to pass and is not inflight or succeeded already
if(workContainer.isDelayPassed()){
    if(workContainer.isNotInFlight() && !workContainer.isUserFunctionSucceeded()){
        count++;
    }
} else {
     // early stop since retryQueue is sorted by retryDueAt
     break;
}
```
The remainder of the queue - those that have their delay not passed yet are OK in that regard as it cannot be inflight if its still waiting for retry delay to pass.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ProcessingShard.java
```diff
@@ -94,10 +103,16 @@ public long getCountWorkInFlight() {
...
+            dcrAvailableWorkContainerCntByDelta(1);
+        }
         return entries.remove(offset);
     }
 
 
+
     // remove staled WorkContainer otherwise when the partition is reassigned, the staled messages will:
     // 1. block the new work containers to be picked and processed
     // 2. will cause the consumer to paused consuming new messages indefinitely
```
rkolesnev, line 118: refactor `removeStaleWorkContainersFromShard` to decrement `availableWorkContainerCnt` similarly to the `public WorkContainer<K, V> remove(long offset)` method above

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ShardManager.java
```diff
@@ -132,9 +139,12 @@ ShardKey computeShardKey(ConsumerRecord<?, ?> wc) {
...
+        // all available container count - (still pending for running retry containers count)
+        // => all_available_count - (retryCnt - all_expired_retry_cnt)
+
         return processingShards.values().stream()
                 .mapToLong(ProcessingShard::getCountOfWorkAwaitingSelection)
-                .sum();
+                .sum() - retryItemCnt.get() + getExpiredItemCntFromRetryQueue();
     }
 
     public boolean workIsWaitingToBeProcessed() {
```
rkolesnev, line 150: refactor to just 
```
public boolean workIsWaitingToBeProcessed() {
    return getNumberOfWorkQueuedInShardsAwaitingSelection() > 0;
}
```
As it needs still to take into account retry queue state for this as well.
in addition remove `workIsWaitingToBeProcessed()` in the ProcessingShard class completely as it cannot be accurate without doing either full scan on all Work Containers in the shard - or needs access to retry queue in the ShardManager. As the method is only used from ShardManager - it is safe to just remove it.

rkolesnev (changes requested):

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ShardManager.java
rkolesnev, line 150: refactor `public boolean workIsWaitingToBeProcessed()` to just 
```
public boolean workIsWaitingToBeProcessed() {
    return getNumberOfWorkQueuedInShardsAwaitingSelection() > 0;
}
```
As it needs still to take into account retry queue state for this as well.

in addition remove `workIsWaitingToBeProcessed()` in the ProcessingShard class completely as it cannot be accurate without doing either full scan on all Work Containers in the shard - or needs access to retry queue in the ShardManager. As the method is only used from ShardManager - it is safe to just remove it.

rkolesnev (changes requested):

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ShardManager.java
sangreal: fixed

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ShardManager.java
```diff
@@ -298,4 +321,18 @@ private void initMetrics() {
         numberOfShardsGauge = pcMetrics.gaugeFromMetricDef(PCMetricsDef.NUMBER_OF_SHARDS,
                 this, shardManager -> shardManager.processingShards.keySet().size());
     }
+
+    // get expired items count from retryQueue
+    private long getExpiredItemCntFromRetryQueue() {
+        long count = 0;
+        for (WorkContainer<?, ?> workContainer : retryQueue) {
+            if (workContainer.getDelayUntilRetryDue().isNegative() || workContainer.getDelayUntilRetryDue().isZero()) {
```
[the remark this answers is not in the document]
sangreal: fixed

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ShardManager.java
sangreal: fixed

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ShardManager.java
sangreal: fixed

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ShardManager.java
sangreal: fixed

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/WorkContainer.java
sangreal: fixed

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ProcessingShard.java
sangreal: fixed

sangreal:
@rkolesnev Thanks for your detailed review. I have fixed according to your comment, except for one. Please help review again

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ShardManager.java
rkolesnev: `// 2. the container has been selected and it is inflight but we already slashed them from availableWorkContainerCnt, so should be counted in`
This is not correct - we do not count inflight containers as available for selection.
The intent of the `getCountOfWorkAwaitingSelection` is to get a number for all work that is ready to be processed - to determine if Consumer should poll for more messages - or there is already enough queued for processing - so that excludes work that is already being processed / inflight.

In original code - the condition was
```
public boolean isAvailableToTakeAsWork() {
        return isNotInFlight() && !isUserFunctionSucceeded() && isDelayPassed();
    }
```
And the updated logic should really reflect the original logic.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ProcessingShard.java
rkolesnev: Similarly in the `ShardManager` class `removeStaleContainers()` - should remove stale containers from the `retryQueue` as well.

rkolesnev (changes requested):

sangreal:
@rkolesnev 

> // 2. the container has been selected and it is inflight but we already slashed them from availableWorkContainerCnt, so should be counted in
This is not correct - we do not count inflight containers as available for selection.
The intent of the getCountOfWorkAwaitingSelection is to get a number for all work that is ready to be processed - to determine if Consumer should poll for more messages - or there is already enough queued for processing - so that excludes work that is already being processed / inflight.

Regarding this,  `getCountOfWorkAwaitingSelection`  indeed do get a number for all work that is ready to be processed and excludes inflight messages in the updated logic. This could be reflected in UT.
I think this pr's goal is 
(1) ensure `getCountOfWorkAwaitingSelection` return the correct number 
(2) performance improvement
(3) makes the code as concise and clean as possible.
But simple copy old logic might not be the final goal. Looking forward to hearing your idea on this one.

rkolesnev:
> @rkolesnev
> 
> > // 2. the container has been selected and it is inflight but we already slashed them from availableWorkContainerCnt, so should be counted in
> > This is not correct - we do not count inflight containers as available for selection.
> > The intent of the getCountOfWorkAwaitingSelection is to get a number for all work that is ready to be processed - to determine if Consumer should poll for more messages - or there is already enough queued for processing - so that excludes work that is already being processed / inflight.
> 
> Regarding this, `getCountOfWorkAwaitingSelection` indeed do get a number for all work that is ready to be processed and excludes inflight messages in the updated logic. This could be reflected in UT. I think this pr's goal is (1) ensure `getCountOfWorkAwaitingSelection` return the correct number (2) performance improvement (3) makes the code as concise and clean as possible. But simple copy old logic might not be the final goal. Looking forward to hearing your idea on this one.

Sure - but we have to keep the logic uniform - we cannot exclude inflight work that was never retried, but include inflight work that is being retried - that would just give a weird number / behaviour that differs based on wether the messages were retried or not.
The check would be done only on WorkContainers that are in retry queue and are ready to be retried to determine if they are in fact still in the queue (thus available to be taken as work) or already in flight / being processed (thus not ready to be taken as work).

The implication can be big enough in edge cases. With maxConcurrency of 1000 - in case external system was unavailable for a brief moment - the retry queue can be over 1000 in size and on recovery of external system - 1000 inflight will skew the number available and potentially keep broker poller paused until they all finish processing - leading to delay with polling for more messages etc - as they are deemed as available work instead of not counted as available work.

rkolesnev:
> Let me have another look at the code - i am thinking if `availableWorkContainerCnt` is already being decremented when we take work inflight - regardless of is it from retry queue or not - so we maybe already accounting for them that way...

Ok - yeah - that is already taken care of by decrementing `availableWorkContainerCnt` in the `ProcessingShard.getWorkIfAvailable(...)` call - when we take work into processing. So you are right - we do not need to exclude them again when counting items in retry queue.

rkolesnev:
Ok - so i am happy enough with it - thank you very much for going back and forth with the PR with me.
Only outstanding bit left is to take care of `retryQueue` in `ShardManager` when removing stale work.

sangreal:
> Ok - so i am happy enough with it - thank you very much for going back and forth with the PR with me. Only outstanding bit left is to take care of `retryQueue` in `ShardManager` when removing stale work.

Thanks for your time for the review! Let me change a bit on my previous pr's code related to stale containers handling and get back to you

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/state/ShardManager.java
```diff
@@ -273,14 +286,13 @@ public List<WorkContainer<K, V>> getWorkIfAvailable(final int requestedMaxWorkTo
...
-        boolean removed = processingShards.values().stream()
+        return processingShards.values().stream()
                 .map(ProcessingShard::removeStaleWorkContainersFromShard)
-                .anyMatch(res -> res.equals(true));
-        if (removed) {
-            log.debug("there are stale work containers removed");
-        }
-        return removed;
+                .flatMap(Collection::stream)
+                .map(retryQueue::remove)
```
sangreal, line 294: remove stale container if existing here

sangreal:
@rkolesnev please help check the updates on the stale container removal for retryQueue, thanks a lot for your review again.

rkolesnev (approved):

rkolesnev:
Hi @sangreal - the PR is ready to be merged - can you please sign the Contributor License Agreement (CLA)? 
You can access it through license/cla status - Details.

sangreal:
> Hi @sangreal - the PR is ready to be merged - can you please sign the Contributor License Agreement (CLA)? You can access it through license/cla status - Details.

@rkolesnev I find it quite weird since I signed it last year already since I already contributed. And when I try to check, I could not signed since it showed I already signed. Please let me know if this is a blocker. If it is, I will revoke this one and try sign again.
image

PR #682, opened 2024-01-23
Make PC message buffer size configurable

rkolesnev:
Make PC message buffer size configurable.
Two configuration options added - specifying message buffer size (which will calculate loading factor and set loading factor to static value) and configurable initial and max loading factors.
* messageBufferSize - takes precedence if specified over load factor options - sets buffer size to approximately specified value by setting load factor to static value calculated as `ceiling ( messageBufferSize / maxConcurrency*batchSize) `
* initialLoadFactor - (default 2) - overrides starting load factor value
* maximumLoadFactor - (default 100) - overrides maximum load factor value

Readme updated to add consideration to tune fetch and buffer size in Partition ordering mode as buffer size can be small by default leading to uneven load per partition and starving processing threads.
Test added to illustrate starved threads scenario and tuning of config to prevent it.

### Checklist

- [X] Documentation (if applicable)
- [x] Changelog

### DIFF
--- a/parallel-consumer-core/src/test/java/io/confluent/parallelconsumer/internal/LimitedDynamicExtraLoadFactor.java
+++ b/parallel-consumer-core/src/test/java/io/confluent/parallelconsumer/internal/LimitedDynamicExtraLoadFactor.java
@@ -1,7 +1,7 @@
 package io.confluent.parallelconsumer.internal;
 
 /*-
- * Copyright (C) 2020-2023 Confluent, Inc.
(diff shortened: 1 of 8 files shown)


parallel-consumer-core/src/test/java/io/confluent/parallelconsumer/internal/LimitedDynamicExtraLoadFactor.java
eddyv, line 1: could remove?

eddyv (approved):
lgtm

PR #719, opened 2024-02-19
Fix the example code for skipping records to work properly

seungy0:
Fixed the example code for skipping records to work properly. Previously, the example code did not work correctly because retriesCount was constantly 0.

### Checklist

- [x] Changelog

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -823,7 +823,7 @@ Implementing a https://github.com/confluentinc/parallel-consumer/issues/196[max 
 
         pc.poll(context -> {
             var consumerRecord = context.getSingleRecord().getConsumerRecord();
-            Long retryCount = retriesCount.computeIfAbsent(consumerRecord, ignore -> 0L);
+            Long retryCount = retriesCount.compute(consumerRecord, (key, oldValue) -> oldValue == null ? 0L : oldValue + 1);
             if (retryCount < maxRetries) {
                 processRecord(consumerRecord);
                 // no exception, so completed - remove from map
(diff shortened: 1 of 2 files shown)


rkolesnev (approved):

README.adoc
```diff
@@ -823,7 +823,7 @@ Implementing a https://github.com/confluentinc/parallel-consumer/issues/196[max
 
         pc.poll(context -> {
             var consumerRecord = context.getSingleRecord().getConsumerRecord();
-            Long retryCount = retriesCount.computeIfAbsent(consumerRecord, ignore -> 0L);
+            Long retryCount = retriesCount.compute(consumerRecord, (key, oldValue) -> oldValue == null ? 0L : oldValue + 1);
```
rkolesnev, line 826: Can you please make this change in CoreApp.java - [here](https://github.com/confluentinc/parallel-consumer/blob/20f8b27a7e62798331cfc0d976218f387ec994c2/parallel-consumer-examples/parallel-consumer-example-core/src/main/java/io/confluent/parallelconsumer/examples/core/CoreApp.java#L146)?

The Readme is populated from java examples using template tags.
To update it - the source java code needs updated - linked above - and then either README Run configuration ran in IntelliJ or `mvn asciidoc-template:build` executed.

Note that executing it shows an error - but its a false error - it just says that it didn't find template / readme in submodules - which is fine as template and readme are only in the top level project.

rkolesnev (changes requested):

README.adoc
seungy0: Sure. Thanks for your reply.

Issue #725, opened 2024-03-04
RecordContext should provide access to lastFailureReason


jvissers:
This is a follow-up of https://github.com/confluentinc/parallel-consumer/discussions/720

It is useful for the `RecordContext` to allow accessing the `lastFailureReason`. , along with the other elements like `numberOfFailedAttempts` ,this way we can actually provide insight into what went wrong - for instance when after a maximum number of failed attempts we need to move the event to a DLQ.

PR #726, opened 2024-03-04
GH-725: RecordContext should provide access to lastFailureReason

jvissers:
Exposes `lastFailureReason` from `Recordcontext`

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -18,6 +18,7 @@ endif::[]
 
 === Improvements
 * improvement: add multiple caches for accelerating available container count calculation （#667）
+* improvement: RecordContext now exposes lastFailureReason (#725)
 
 
 == 0.5.2.8
--- a/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/RecordContext.java
+++ b/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/RecordContext.java
@@ -71,5 +71,10 @@ public class RecordContext<K, V> {
     public Optional<Instant> getSucceededAt() {
         return workContainer.getSucceededAt();
     }
+
+    /**
+     * @return if the record has failed, returns the last failure reason
+     */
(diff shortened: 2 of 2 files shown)


jvissers:
I have signed the agreement.

rkolesnev:
Hi @jvissers - is there a mismatch between email configured on your GH client / to do the commit and configured on your GH account? I think CLA bot is cross-checking on that - if they don't match - that may trip it over.
Can you please add the change to the CHANGELOG.adoc - it is a small change - but i think it is a useful one - so would be good to highlight in the changelog.

Thanks.

rkolesnev:
Here is a GH troubleshooting link for account / local account setup - if that helps [troubleshooting-commits](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/troubleshooting-commits/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user)
Looking at the commit here - your username is not resolved as a link to GH user - there may be something missing in your local account setup - potentially that prevents CLA bot from working properly - as it cant resolve author on the commit to your GH user.

jvissers:
I've updated changelog, updated the git config for this project locally and
pushed my changes.

Apparently still not everything is working as desired. I've followed the
link for troubleshooting-commits, but that doesn't seem to reflect what I'm
seeing.

For one this ("over blue ? to the right of the username") as gets mentioned
here: "To read a message about why the commit is not linked, hover over the
blue?  to the right of the username.", is not showing. What else can I do?

rkolesnev:
could you please try to squash the commits on your branch or reset and recommit and then force push the changes?

Basically the first commit still has your oracle email associated with it - the local changes are not retrospective so they only applied to the 2nd commit (change to the changelog).

I am hoping that force-push with new commit(s) that have updated email - will let the check to re-evaluate properly.
Alternatively - new PR would need to be created - and should go through fine as long as commits in it are using your Gmail email.

Thanks.

Issue #721, opened 2024-02-27
`checkAutoCommitIsDisabled` fails for `kafka-clients` v3.7.0


nscuro:
When using PC with `kafka-clients` v3.7.0, PC fails while trying to determine whether the provided `Consumer` has auto-commit enabled:

```
Caused by: java.lang.IllegalStateException: Cannot check auto commit is disabled for consumer type: org.apache.kafka.clients.consumer.KafkaConsumer
	at io.confluent.parallelconsumer.internal.AbstractParallelEoSStreamProcessor.checkAutoCommitIsDisabled(AbstractParallelEoSStreamProcessor.java:493)
	at io.confluent.parallelconsumer.internal.AbstractParallelEoSStreamProcessor.validateConfiguration(AbstractParallelEoSStreamProcessor.java:330)
	at io.confluent.parallelconsumer.internal.AbstractParallelEoSStreamProcessor.<init>(AbstractParallelEoSStreamProcessor.java:283)
	at io.confluent.parallelconsumer.internal.AbstractParallelEoSStreamProcessor.<init>(AbstractParallelEoSStreamProcessor.java:266)
	at io.confluent.parallelconsumer.ParallelEoSStreamProcessor.<init>(ParallelEoSStreamProcessor.java:46)
	at io.confluent.parallelconsumer.ParallelStreamProcessor.createEosStreamProcessor(ParallelStreamProcessor.java:26)
...
Caused by: java.lang.NoSuchFieldException: coordinator
	at java.base/java.lang.Class.getDeclaredField(Class.java:2610)
	at io.confluent.parallelconsumer.internal.AbstractParallelEoSStreamProcessor.checkAutoCommitIsDisabled(AbstractParallelEoSStreamProcessor.java:473)
```

Upon surface-level investigation, the failure is caused by https://issues.apache.org/jira/browse/KAFKA-15277. `KafkaConsumer` no longer holds a `coordinator` field. 

Instead, `KafkaConsumer` now has a `ConsumerDelegate<K, V> delegate` field, which resolves to either `LegacyKafkaConsumer`, or `AsyncKafkaConsumer`.

`LegacyKafkaConsumer` [still has a `coordinator` field](https://github.com/apache/kafka/blob/3.7.0/clients/src/main/java/org/apache/kafka/clients/consumer/internals/LegacyKafkaConsumer.java#L118), but `AsyncKafkaConsumer` does not.

rkolesnev:
[...]

PR #746, opened 2024-03-30
Kafka 3.7.0 fixes

ondryaso:
This is the bare minimum to make the library work with version 3.7.0 of the Kafka client. It fixes the bug reported in #721 and also introduces the option, mentioned therein, to disable the auto commit check (by making it "soft", i.e. not throwing an exception).

### Checklist

- [ ] Changelog


parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumerOptions.java
```diff
@@ -535,4 +535,12 @@ public boolean isProducerSupplied() {
...
     @Builder.Default
     public final int maximumLoadFactor = DynamicLoadFactor.DEFAULT_MAX_LOADING_FACTOR;
+
+    /**
+     * Make the "auto commit is disabled check" soft - when set, the processor will not throw an exception if Kafka's
+     * auto commit feature is enabled in the consumer. Use with caution, the library coordinates offset so
+     * <code>enable.auto.commit</code> must always be disabled in your consumer.
+     */
+    @Builder.Default
+    public final boolean softAutoCommitDisabledCheck = false;
```
rkolesnev, line 545: I think the name can be more explicit.
```suggestion
    public final boolean ignoreReflectiveAccessExceptionsForAutoCommitDisabledCheck = false;
```

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/ParallelConsumerOptions.java
```diff
@@ -535,4 +535,12 @@ public boolean isProducerSupplied() {
      */
     @Builder.Default
     public final int maximumLoadFactor = DynamicLoadFactor.DEFAULT_MAX_LOADING_FACTOR;
+
+    /**
+     * Make the "auto commit is disabled check" soft - when set, the processor will not throw an exception if Kafka's
+     * auto commit feature is enabled in the consumer. Use with caution, the library coordinates offset so
+     * <code>enable.auto.commit</code> must always be disabled in your consumer.
```
rkolesnev, line 542:
```suggestion
     * The purpose of the flag is to be a last resort / temporary work-around for changes introduced in newer Kafka
     * Clients that break reflective access and for using wrapped, custom or extended KafkaConsumer classes that fail
     * reflection checks - setting the flag to true will ignore reflection access exceptions during this check.
     * <p>
     * Note: that library will still try to access auto commit field on the consumer object and if it is accessible and
     * not disabled - the Parallel Consumer will shut down.
```

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -470,27 +473,50 @@ private void checkAutoCommitIsDisabled(org.apache.kafka.clients.consumer.Consume
...
 
-                if (TRUE.equals(isAutoCommitEnabled))
-                    throw new ParallelConsumerException("Consumer auto commit must be disabled, as commits are handled by the library.");
+                if (TRUE.equals(isAutoCommitEnabled)) {
+                    if (options.isSoftAutoCommitDisabledCheck()) {
+                        log.warn("Consumer auto commit must be disabled, as commits are handled by the library.");
+                    } else {
+                        throw new ParallelConsumerException("Consumer auto commit must be disabled, as commits are handled by the library.");
+                    }
+                }
```
rkolesnev, line 507: Revert this check back - only disabling error on reflective access - if auto commit field is accessible / found - and is not disabled - shut down with an error.
```suggestion
                if (TRUE.equals(isAutoCommitEnabled))
                    throw new ParallelConsumerException("Consumer auto commit must be disabled, as commits are handled by the library.");
```

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -470,27 +473,50 @@ private void checkAutoCommitIsDisabled(org.apache.kafka.clients.consumer.Consume
...
+                    Field autoCommitEnabledField = coordinator.getClass().getDeclaredField("autoCommitEnabled");
+                    autoCommitEnabledField.setAccessible(true);
+                    isAutoCommitEnabled = (Boolean) autoCommitEnabledField.get(coordinator);
+                } else if (delegate instanceof AsyncKafkaConsumer) {
+                    Field autoCommitEnabledField = AsyncKafkaConsumer.class.getDeclaredField("autoCommitEnabled");
+                    autoCommitEnabledField.setAccessible(true);
+                    isAutoCommitEnabled = (Boolean) autoCommitEnabledField.get(delegate);
+                } else {
+                    log.error("Consumer delegate is neither a LegacyKafkaConsumer nor a AsyncKafkaConsumer - cannot check auto commit is disabled for consumer type: " + consumer.getClass().getName());
+                    isAutoCommitEnabled = false;
```
rkolesnev, line 498: Check the new option flag and either throw an error or log a warning depending on how that flag is set - something like below:
```suggestion
               if (options.ignoreReflectiveAccessExceptionsForAutoCommitDisabledCheck()) {
                    log.warn("Consumer delegate is neither a LegacyKafkaConsumer nor a AsyncKafkaConsumer - cannot check auto commit is disabled for consumer type: {}. Ignoring due to the option flag ignoreReflectiveAccessExceptionsForAutoCommitDisabledCheck=true", consumer.getClass().getName());
                } else {
                    throw new ParallelConsumerException("Consumer delegate is neither a LegacyKafkaConsumer nor a AsyncKafkaConsumer - cannot check auto commit is disabled for consumer type: " + consumer.getClass().getName());
                }
```

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -470,27 +473,50 @@ private void checkAutoCommitIsDisabled(org.apache.kafka.clients.consumer.Consume
...
                 log.debug("Detected MockConsumer class which doesn't do auto commits");
             } else {
                 // Probably Mockito
                 log.error("Consumer is neither a KafkaConsumer nor a MockConsumer - cannot check auto commit is disabled for consumer type: " + consumer.getClass().getName());
             }
-        } catch (NoSuchFieldException | IllegalAccessException e) {
-            throw new IllegalStateException("Cannot check auto commit is disabled for consumer type: " + consumer.getClass().getName(), e);
+        } catch (NoSuchFieldException | IllegalAccessException | NullPointerException e) {
+            if (options.isSoftAutoCommitDisabledCheck()) {
+                log.error("Cannot check auto commit is disabled for consumer type: " + consumer.getClass().getName(), e);
```
rkolesnev, line 516: add flag name to make it clear that its only allowed due to the explicit override flag:
```suggestion
                log.error("Cannot check auto commit is disabled for consumer type: {},  Ignoring due to the option flag ignoreReflectiveAccessExceptionsForAutoCommitDisabledCheck=true", consumer.getClass().getName(), e);
```

rkolesnev (changes requested):
Overall looks good. 
I have left some comments - mainly to make the new flag more explicit in that it only allows to suppress reflective access errors - not disable the auto commit check itself.
I dont see any reason to disable auto commit check - apart from failures to actually perform the check.

Another small thing - can you please update the Changelog and regenerate Readme - to regenerate readme either run the maven target in IDE `README` or `mvn asciidoc-template:build` - ignore errors on submodules - as readme is only on top level - it reports missing template on submodules as errors.

nscuro:
Hey @ondryaso, thanks so much for taking this up! Are you planning to address the requested changes?

I can try picking up from here in case other commitments prevent you from continuing :)

ondryaso:
@nscuro Hey, that would be awesome, I am currently totally out of time...

nscuro:
With a bit of a delay, I now raised #762 as a replacement for this PR. Thanks for your work @ondryaso, I think you can close this one.

ondryaso:
Thanks, @nscuro, closing this PR.

PR #779, opened 2024-05-30
stale containers exclusion and handling improvement

sangreal:
Description...
in previous logic we need two passes to get the normal containers and put stale containers into mailbox queue.
But now we only need one pass, therefore when the traffic is huge, there will be performance gain.
### Checklist

- [x] Changelog

### DIFF
--- a/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
+++ b/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
@@ -139,7 +139,7 @@ public abstract class AbstractParallelEoSStreamProcessor<K, V> implements Parall
      */
     @Value
     @RequiredArgsConstructor(access = PRIVATE)
-    private static class ControllerEventMessage<K, V> {
+    static class ControllerEventMessage<K, V> {
 
         WorkContainer<K, V> workContainer;
 
@@ -1285,7 +1285,12 @@ public abstract class AbstractParallelEoSStreamProcessor<K, V> implements Parall
          *  Handle and filter stale work from the batch, before creating the internal context for running user function.
          *  The context created is used by the "wrapped" user function to inject transactional producer synchronization.
          */
-        final List<WorkContainer<K, V>> activeWorkContainers = handleStaleWork(workContainerBatch);
+        Map<Boolean, List<WorkContainer<K, V>>> splitContainersMap = workContainerBatch.stream()
+                .collect(Collectors.groupingBy(wm::checkIfWorkIsStale));
+        final List<WorkContainer<K, V>> staleWorkContainers = splitContainersMap.getOrDefault(Boolean.TRUE, new ArrayList<>());
+        final List<WorkContainer<K, V>> activeWorkContainers = splitContainersMap.getOrDefault(Boolean.FALSE, new ArrayList<>());
+
+        handleStaleWork(staleWorkContainers);
 
         final PollContextInternal<K, V> context = new PollContextInternal<>(activeWorkContainers);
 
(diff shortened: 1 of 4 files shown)


rkolesnev:
Hey @sangreal - is the added complexity worth the improvement?
My understanding that performance improvement will only apply when there is stale work due to rebalance / epoch change - but such an event should really be rare relatively speaking - so few millisecond performance improvement on one-off scans wouldn't warrant additional changes in my opinion.
What do you think?

sangreal:
@rkolesnev Thanks for the reply.
1.  I think the probability is not that low since when I encountered the staled containers causing poller paused last year, the assigned to the same partition causing stale containers happened a time a few days using `cooperativestickyassignor`.
Therefore the fix could be valuable.
2. the new logic is quite straightforward and with UTs, I don't think this is quite complex.

Happy to hear your opinion on these.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
```diff
@@ -1287,15 +1287,8 @@ protected <R> List<ParallelConsumer.Tuple<ConsumerRecord<K, V>, R>> runUserFunct
          */
         final boolean containsStaleWork = wm.checkIfWorkIsStale(workContainerBatch);
```
rkolesnev, line 1288: ` wm.checkIfWorkIsStale(workContainerBatch)` is iterating over each work container and checking if it is stale - so it feels like redundant to do if we are trying to eliminate the re-scanning / re-iterating over them.

I think the filtering can happen here straight away - as we are already iterating over them anyway.
Basically the logic would look like:
 ``` 
  final Map<Boolean, List<WorkContainer<K, V>>> splitContainersMap = workContainerBatch.stream()
                .collect(Collectors.groupingBy(wm::checkIfWorkIsStale));
  final List<WorkContainer<K, V>> staleWorkContainers = splitContainersMap.getOrDefault(Boolean.TRUE, new ArrayList<>());
  if(!staleWorkContainers.isEmpty()){
      handleStaleWork(staleWorkContainers);
  }
  final List<WorkContainer<K, V>> activeWorkContainers = splitContainersMap.getOrDefault(Boolean.FALSE, new ArrayList<>());
...

```
`handleStaleWork` then accepts already filtered, only stale containers and we have `activeWorkContainers` pre-filtered and ready for processing further down.

We avoid having `handleStaleWork` method that performs hidden side effect (filtering out of stale work containers) and logic is easier to follow at a glance.

**Review by rkolesnev — CHANGES_REQUESTED** [review]

**Comment by rkolesnev** [comment]
And if you can please run license:format or update license headers in the changed test files to 2024

File: parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java (hunk 1 above)
Comment (sangreal) [reply parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java:1288]: Thanks a lot. I think this is very good idea, I miss this in the first place. Let me simplify this logic.

Comment (sangreal) [reply parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java]: fixed

**Comment by sangreal** [comment]
> And if you can please run license:format or update license headers in the changed test files to 2024

Sure, headers have been updated

File: parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java [hunk 2]
```diff
@@ -1331,19 +1320,22 @@ protected <R> List<ParallelConsumer.Tuple<ConsumerRecord<K, V>, R>> runUserFunct
...
      */
-    protected void handleStaleWork(final List<WorkContainer<K, V>> workContainerBatch) {
-        final List<WorkContainer<K, V>> staleWorkContainers = workContainerBatch
-                .stream()
-                .filter(wm::checkIfWorkIsStale)
-                .collect(Collectors.toList());
+    protected List<WorkContainer<K, V>> handleStaleWork(final List<WorkContainer<K, V>> workContainerBatch) {
+        Map<Boolean, List<WorkContainer<K, V>>> splitContainersMap = workContainerBatch.stream()
+                .collect(Collectors.groupingBy(wm::checkIfWorkIsStale));
+        final List<WorkContainer<K, V>> staleWorkContainers = splitContainersMap.getOrDefault(Boolean.TRUE, new ArrayList<>());
```
rkolesnev, line 1326: Could you please move this up into runUserFunction - to line 1288.
I would prefer for inner functions to have single purpose - so handleStaleWork only handles stale work and doesn't have additional / semi-hidden logic to filter active containers in it...

rkolesnev (changes requested):

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/AbstractParallelEoSStreamProcessor.java
sangreal: Thanks for the comments. I addressed your comment and update UT accordingly.
Please have a look when you have time.

rkolesnev (approved):

PR #812, opened 2024-07-08
simplify partition and fix typo

sangreal:
Description...
1. simplify partition method
2. fix typo

### Checklist

- [x] Changelog

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -14,10 +14,15 @@ ifndef::github_name[]
 toc::[]
 endif::[]
 
+== 0.5.3.1
+
+=== Fixes
+* fix: simplify partition and fix typo (#812)
+
+
 == 0.5.3.0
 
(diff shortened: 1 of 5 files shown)


rkolesnev:
Hey @sangreal - looks good to me in general - but is there any optimization in terms of performance gained here? 
Otherwise i, personally, find it easier to read the old code that does partitioning into batches step by step - rather than working out how exactly the sublisting / chunking lambda works.
If there is no benefit in terms of performance (or it is really insignificant) - i think it is better to keep it as is.

sangreal:
@rkolesnev , I don't think there will be any performance gain but only code concise improvement. 
I am very open to revert this parts changes, if you think typo fixes is worthwhile, I could revert other changes but only leave typo fixes.

sangreal:
close this since this pr is not meaningful now.

PR #816, opened 2024-07-25
Prevent commitSync to bubble up exception to control thread where no exception is caught.

wushilin:
Moved commitSync to a while loop to attempt to commit forever. As soon as brokers are online again, commit will succeed and consumption can resume.

### Checklist

- [x] Changelog

### DIFF
--- a/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/ConsumerManager.java
+++ b/parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/ConsumerManager.java
@@ -94,7 +94,21 @@ public class ConsumerManager<K, V> {
         noWakeups++;
         while (inProgress) {
             try {
-                consumer.commitSync(offsetsToSend);
+                while(true) {
+                    try {
+                        consumer.commitSync(offsetsToSend);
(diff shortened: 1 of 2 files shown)


parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/ConsumerManager.java
```diff
@@ -94,7 +94,21 @@ public void commitSync(final Map<TopicPartition, OffsetAndMetadata> offsetsToSen
         noWakeups++;
         while (inProgress) {
             try {
-                consumer.commitSync(offsetsToSend);
+                while(true) {
+                    try {
+                        consumer.commitSync(offsetsToSend);
+                        // break when offset commit is okay. Do not throw exception to main threads
+                        break;
+                    } catch(Throwable t) {
```
nscuro, line 102: Wouldn't it be better to explicitly specify the exceptions that are retriable? You wouldn't want to be stuck in a commit loop if the call has no chance of ever succeeding.

rkolesnev:
@wushilin - can you have a look at #818  - i know it is solving this problem in a different manner from your suggested approach - but i think it may fix it for your use case as well - and i am in favour of it to be honest.
Closing this PR.

PR #819, opened 2024-08-15
Improved offset commit retry. Add support for SaslAuthenticationException retry timeout

wushilin:
In this patch, we changed ConsumerManager.commitSync to honour the user set offsetCommitTimeout value.

We also added SaslAuthenticationException retry timeout to support (`saslAuthenticationRetryTimeout`) retries up to user configured value.

The code should not break any existing configuration or default configuration.
When saslAuthenticationRetryTimeout and offsetCommitTimeout are set to high enough (e.g. 30 minutes)
Based on test, it can survive whole cluster down, or temporary LDAP down and will recover automatically upon broker restart or LDAP restart.

Note:
If user use default configuration, there is almost no difference because default offset commit timeout is 30 seconds and Kafka Consumer commit timeout is actually 60 seconds. Therefore no retry will be performed.

If user use default saslAuthenticationRetryTimeout of 0, SaslAuthenticationException will be bubbled up immediately without retry.

### Checklist

- [ x ] Documentation (if applicable) - Javadoc had been added
- [ x ] Changelog


rkolesnev:
How would it behave if you try to close parallel consumer while its retrying poll due to SASLAuth error or commit ? would those retries block the closing down process? If so i would add a flag like AtomicBoolean closeRequested to the ConsumerManager - default it to false, set it to true in the close() method and add it into consideration if retry should be performed - then would need to sequence - stop retrying, then close consumer.

wushilin:
> How would it behave if you try to close parallel consumer while its retrying poll due to SASLAuth error or commit ? would those retries block the closing down process? If so i would add a flag like AtomicBoolean closeRequested to the ConsumerManager - default it to false, set it to true in the close() method and add it into consideration if retry should be performed - then would need to sequence - stop retrying, then close consumer.

Makes sense, let me enhance this logic, and also add test together.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/ConsumerManager.java
```diff
@@ -55,7 +67,36 @@ ConsumerRecords<K, V> poll(Duration requestedLongPollTimeout) {
...
+                        records = consumer.poll(timeoutToUse);
+                        break;
+                    } catch (SaslAuthenticationException authenticationException) {
+                        Instant now = Instant.now();
+                        Duration elapsed = Duration.between(pollStarted, now);
+                        boolean shouldRetry = elapsed.toMillis() < saslAuthenticationRetryTimeout.toMillis();
+                        if (shouldRetry) {
+                            log.warn("Poll error: SaslAuthenticationException. Retrying ({})", tryCount);
+                            try {
+                                retryBackOff(5000L); // no need to check return value here as next loop will check
```
rkolesnev, line 86: can you please change this to constant and reduce to 1 second - so that timeout can be set a bit more granular.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/ConsumerManager.java
```diff
@@ -94,15 +135,82 @@ public void commitSync(final Map<TopicPartition, OffsetAndMetadata> offsetsToSen
...
+                        // We should honor the user configured SaslAuthenticationException timeout here.
+                        // to allow the program to sustain temporary LDAP failures
+                        Instant now = Instant.now();
+                        Duration elapsed = Duration.between(startedTime, now);
+                        boolean shouldRetry = elapsed.toMillis() <= saslAuthenticationRetryTimeout.toMillis();
+                        if(shouldRetry) {
+                            log.warn("Encountered SaslAuthenticationException while committing offset. Retrying ({})", tryCount);
+                            // Since authentication exception may happen immediately, it is good to sleep a few seconds before trying again
+                            try {
+                                retryBackOff(5000L); // no need to check return value
```
rkolesnev, line 176: same - reuse common constant in both places for Sasl backoff

parallel-consumer-core/src/test/java/io/confluent/parallelconsumer/MockConsumerTestWithCommitTimeoutException.java
```diff
@@ -0,0 +1,127 @@
...
+ * <p>
+ * These tests demonstrate why using {@link MockConsumer} is difficult, and why {@link LongPollingMockConsumer} should
+ * be used instead.
+ *
+ * @author Antony Stubbs
+ * @see LongPollingMockConsumer#revokeAssignment
+ */
+@Slf4j
+@Timeout(60000L)
+class MockConsumerTestWithCommitTimeoutException {
```
rkolesnev, line 42: Update the comment above to be relevant to the new test.
Update / add comments in the test explaining what it is doing / expecting etc.

parallel-consumer-core/src/test/java/io/confluent/parallelconsumer/MockConsumerTestWithEarlyClose.java
```diff
@@ -0,0 +1,126 @@
...
+ * <p>
+ * These tests demonstrate why using {@link MockConsumer} is difficult, and why {@link LongPollingMockConsumer} should
+ * be used instead.
+ *
+ * @author Antony Stubbs
+ * @see LongPollingMockConsumer#revokeAssignment
+ */
+@Slf4j
+@Timeout(60000L)
+class MockConsumerTestWithEarlyClose {
```
rkolesnev, line 40: Update the comment above to be relevant to the new test.
Update / add comments in the test explaining what it is doing / expecting etc.

parallel-consumer-core/src/test/java/io/confluent/parallelconsumer/MockConsumerTestWithSaslAuthenticationException.java
```diff
@@ -0,0 +1,114 @@
...
+ * <p>
+ * These tests demonstrate why using {@link MockConsumer} is difficult, and why {@link LongPollingMockConsumer} should
+ * be used instead.
+ *
+ * @author Antony Stubbs
+ * @see LongPollingMockConsumer#revokeAssignment
+ */
+@Slf4j
+@Timeout(60000L)
+class MockConsumerTestWithSaslAuthenticationException {
```
rkolesnev, line 41: Update the comment above to be relevant to the new test.
Update / add comments in the test explaining what it is doing / expecting etc.

rkolesnev (changes requested):

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/ConsumerManager.java
wushilin: Will make it a constant for retry backoff, but the retryBackOff is already checking the shutdownRequested flag in 100ms intervals. It will return early so the while loop will break by itself as early as possible.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/BrokerPollSystem.java
```diff
@@ -266,6 +267,7 @@ private void doPause() {
     }
 
     public void closeAndWait() throws TimeoutException, ExecutionException {
+        consumerManager.signalStop();
```
rkolesnev, line 270: redundant - transitionToClosing() method does that signalling already.

parallel-consumer-core/src/main/java/io/confluent/parallelconsumer/internal/ConsumerManager.java
```diff
@@ -45,7 +59,7 @@ public class ConsumerManager<K, V> {
 
     ConsumerRecords<K, V> poll(Duration requestedLongPollTimeout) {
         Duration timeoutToUse = requestedLongPollTimeout;
-        ConsumerRecords<K, V> records;
+        ConsumerRecords<K, V> records = new ConsumerRecords<>(new HashMap<>());
```
rkolesnev, line 62: I think can just init to null avoiding reduntant hashmap / ConsumerRecords object creation and do a check on return
```suggestion
        ConsumerRecords<K, V> records = null;
```

Then:
```
Line 112: return records;
```
replaced with 
 ```
 return records != null ? records : new ConsumerRecords<>(UniMaps.of());
 ```
 and due to that change can remove statement in the catch block on  Line 108: ` records = new ConsumerRecords<>(UniMaps.of());`

rkolesnev (changes requested):
couple of small changes.
Everything else looks good. thank you

rkolesnev (approved):

eddyv (approved):

niamhthornbury (approved):

Issue #896, opened 2025-10-27
Add support for virtual threads


bdeneuter:
Hi,

This is a feature request. 
Virtual threads were added to the JDK in JDK 21. It still had an issue with pinning which has been fixed with https://openjdk.org/jeps/491 in JDK 24.
JDK 25 became the first LTS which has support for virtual threads without the pinning issue.
It would be nice to have support for virtual threads for processing when running on JDK 25.
I created a fork with a first implementation but I didn't find the time to get the Maven build working correctly. 
I tried to keep the original build and have only the module parallel-consumer-virtual-threads build on JDK 25 but didn't succeed yet.
You find a rough version from the code here:
https://github.com/bdeneuter/parallel-consumer/blob/master/parallel-consumer-virtual-threads/src/main/java/io/confluent/parallelconsumer/virtual_thread/VirtualThreadProcessor.java

I based the implementation on the reactor version.

mageddo:
I achieved virtual threads with   `io.confluent.parallelconsumer:parallel-consumer-core:0.5.3.3` by doing the following:

```java
public class Main {

  static {
    new VirtualThreadJavaxNamingConfigurer().configure();
  }

   public static void main(String[] args) {
     ..
  }

}
```


```java
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.ThreadFactory;

import javax.naming.Context;

import com.mageddo.naming.Contexts;
import com.mageddo.naming.SimpleContextFactory;

import lombok.extern.slf4j.Slf4j;

@Slf4j
public class VirtualThreadJavaxNamingConfigurer {

  /**
   * Path to Managed executor service for Java EE
   */
  private static final String MANAGED_EXECUTOR_SERVICE = "java:comp/DefaultManagedExecutorService";

  /**
   * Path to Managed thread factory for Java EE
   */
  private static final String MANAGED_THREAD_FACTORY = "java:comp/DefaultManagedThreadFactory";

  private final ExecutorService executorService;
  private final ThreadFactory threadFactory;

  public VirtualThreadJavaxNamingConfigurer() {
    this.executorService = Executors.newVirtualThreadPerTaskExecutor();
    this.threadFactory = Thread.ofVirtual()
        .name("virt-", 0)
        .factory();
  }

  public void configure() {
    System.setProperty(Context.INITIAL_CONTEXT_FACTORY, SimpleContextFactory.class.getName());
[...]

Issue #862, opened 2025-04-08
Parallel Consumer cannot run on Java 24


jvissers:
This project doesn't owe me/us anything, but I feel at this point it is justified to raise an issue for this. 

Effectively this is also the subject of this discussion https://github.com/confluentinc/parallel-consumer/discussions/848 - which hasn't seen any traction yet. Essentially what this probably boils down to, is whether Parallel Consumer will be able to run on Kafka Client 3.9.1 - when that becomes available. That version will, for one, have the removal of the Security Manager (See, https://github.com/apache/kafka/pull/19221 and https://github.com/apache/kafka/pull/19368).

The fact that parallel consumer, now only uses Kafka 3.7.1 is already kind of a problem - as 3.9.0 has been out there for quite some time. At present, we are using Kafka clients 3.8.1 as a 4th party dependency through Helidon MP 4.1.6.

So the bottom-line is:

- parallel consumer should probably bump its dependencies to Kafka client version 3.9.0
- the moment Kafka client version 3.9.1 becomes available, we'd want to be able to use that version - and have support for Java 24.

astubbs:
Tracked in the maintained fork as astubbs/parallel-consumer#181, which carries a summary of this
issue and its current status there.

The **Confluent-hosted version** of this project is no longer maintained - its README says so.
**Parallel Consumer itself continues** in my fork: https://github.com/astubbs/parallel-consumer

Nothing in this repository is closed or changed by this comment.

astubbs:
@jvissers, @ajax-surovskyi-y - the blocker this report named is gone in the maintained fork's [v0.6.0.0](https://github.com/astubbs/parallel-consumer/releases/tag/v0.6.0.0), which ships on kafka-clients 3.9.2. The coordinates changed with the fork - on Maven Central it is now:

```xml
<dependency>
    <groupId>bz.stub.parallelconsumer</groupId>
    <artifactId>parallel-consumer-core</artifactId>
    <version>0.6.0.0</version>
</dependency>
```

(`parallel-consumer-vertx`, `parallel-consumer-reactor` and `parallel-consumer-mutiny` likewise.) The Java package is `bz.stub.parallelconsumer` too, so imports change with it - one command rewrites them (GNU `sed`; on macOS use `sed -i ''`):

```bash
find . -name '*.java' -exec sed -i 's/io\.confluent\.parallelconsumer/bz.stub.parallelconsumer/g' {} +
```

[...]

PR #908, opened 2026-01-12
[Feature] Support Virtual Threads

devingryu:
Adds option to use Virtual Threads for worker pool. (See also: #896)

* VT support: Added the `useVirtualThreads` option and execution logic to support Java 21+ Virtual Threads (Project Loom).
* VT Pinning prevention: Replace synchronization method from `synchronized` blocks to `ReentrantLock` to prevent VT pinning.
* ExecutorService Abstraction: Generalized the return type of `setupWorkerPool` to `ExecutorService` and improved queue management and backpressure logic accordingly.

### Concerns

1.  Test Environment Compatibility
    The newly added VT-related tests require JDK 21+ and are configured to skip execution using `Assumptions` if the environment does not support it. It seems that current CI environment is based on JDK 17, if so the tests I added will be skipped in CI.

2.  Concurrency in `isPoolQueueLow`
    When VT mode is enabled, we cannot directly query the queue size, so `WorkerManager.numberRecordsOutForProcessing` is used to calculate the load. Please review if this approach creates any potential concurrency issues in a multi-threaded environment.

3.  Defensive Logic
    To ensure the stability of the existing logic, if VT is disabled (`!useVirtualThreads`), `setupWorkerPool` is enforced to return a `ThreadPoolExecutor` (throwing an exception otherwise), and type checks have been added. I am concerned this might introduce a breaking change (compatibility issue) for existing users who inherit `AbstractParallelEoSStreamProcessor` to use custom Executors.

### Checklist

- [x] Changelog

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -16,6 +16,14 @@ endif::[]
 
 == 0.5.3.4
 
+=== Features
+
+* feature: add options to enable virtual threads
+
+=== Improvements
+
+* refactor: migrate synchronized into ReentrantLock
+
 === Fixes
 
 * fix: keep instantiated OffsetMapCodecManager so that metrics will not be recreated every commit (#859)
--- a/README.adoc
+++ b/README.adoc
@@ -1536,6 +1536,14 @@ endif::[]
 
 == 0.5.3.4
 
+=== Features
+
+* feature: add options to enable virtual threads
+
+=== Improvements
+
+* refactor: migrate synchronized into ReentrantLock
+
 === Fixes
 
 * fix: keep instantiated OffsetMapCodecManager so that metrics will not be recreated every commit (#859)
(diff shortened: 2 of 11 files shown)


Add0z:
Is someone coming to approve these PRs?

astubbs:
@devingryu - the maintained fork went a different way on virtual threads, and it is fair to say so on this PR rather than only on the issue: astubbs/parallel-consumer#360 runs the user function on virtual threads as an opt-in on JDK 21 and later, with the worker pool reshaped for it, rather than migrating the `synchronized` blocks. It is listed under What comes next in the [v0.6.0.0](https://github.com/astubbs/parallel-consumer/releases/tag/v0.6.0.0) notes and is not yet merged. This PR's `ReentrantLock` migration was also examined for the Java 24 question in confluentinc/parallel-consumer#862 and is a different mechanism from that one. Thank you for the work; the fork is where the work continues.

```xml
<dependency>
    <version>0.6.0.0</version>
</dependency>
```

```bash
```

